PR #2, opened 2016-02-11
Fix/spark 1.6

hkothari:
Upgrades to Spark 1.6 without the need for using internal row. This allows us to keep SchemaConverters.scala the same and eventually use the one in spark-avro that has now been made public.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
@@ -19,28 +19,18 @@ package org.apache.spark.sql.execution.datasources.hbase
 
 import java.util.ArrayList
 
+import org.apache.hadoop.hbase.CellUtil
+import org.apache.hadoop.hbase.client._
 import org.apache.hadoop.hbase.filter.{Filter => HFilter, FilterList => HFilterList}
 import org.apache.hadoop.hbase.util.Bytes
-import org.apache.spark.sql.catalyst.expressions.Row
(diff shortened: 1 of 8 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
```diff
@@ -93,34 +82,34 @@ private[hbase] class HBaseTableScanRDD(
...
-          row.setNullAt(x._2)
+          null
         } else {
           val v = CellUtil.cloneValue(kv)
-          Utils.setRowCol(row, x, v, 0, v.length)
+          Utils.hbaseFieldToScalaType(x, v, 0, v.length)
         }
       }
     }
+    Row.fromSeq(valueSeq)
```
zhzhan: Potentially it will create a lot of small row object, increase gc overhead. But will commit it because it looks more elegant in the surface.

hkothari: It actually shouldn't behave too differently from the MutableRow approach because we still only create one row object per corresponding HBase row. The difference is that we create a Seq before hand in this version and in the previous version it created a mutable row and then edited that mutable row.

In either case it's one Row object per HBase row, this one just has one addition of a Seq before that..

zhzhan: In previous version, the mutable row is reused, but in this one row is created for each hbase row.

hkothari: Ahh, good catch!

PR #3, opened 2016-02-26
Variable length key partial support.

hkothari:
This is partial support because it doesn't handle the situation where you have variable keys inbetween fixed length keys.

For example: `string(6):string(?):int` would not work because we don't actually write delimiters with the variable length string.

To support this, we need to add the `HBaseTableCatalog.delimiter` after the end of variable length strings. Parsing this into a row object is already supported with the current code, but the filter logic in HBaseFilter.scala breaks because we need to disable push down filters on keys which are after the fixed length portion (ie. the int column in the above example).

Altogether the logic in HBaseFilter is kind of sketchy because it treats LessThan and LessThanEqualTo (as well as the GreaterThan counterpart) as the same, so we should probably rethink that.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
@@ -29,7 +29,7 @@ import org.apache.spark.sql.Row
 import org.apache.spark.sql.execution.datasources.hbase
 import org.apache.spark.sql.execution.datasources.hbase.HBaseResources._
 import org.apache.spark.sql.sources.Filter
-import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.types.{StringType, StructType}
 
 import scala.collection.mutable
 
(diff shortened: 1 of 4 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
```diff
@@ -81,35 +81,54 @@ private[hbase] class HBaseTableScanRDD(
...
+      if (kv == null || kv.getValueLength == 0) {
+        (x, null)
+      } else {
+        val v = CellUtil.cloneValue(kv)
+        (x, Utils.hbaseFieldToScalaType(x, v, 0, v.length))
+      }
+    }.toMap
+    val unioned = keySeq ++ valueSeq
+    // Return the row ordered by the requested order
+    Row.fromSeq(fields.map(unioned.get(_)))
```
zhzhan: Need to change to Row.fromSeq(fields.map(unioned.get(_).getOrElse(null)))

PR #12, opened 2016-06-21
Suggested update to attribute trademarks

ShaneCurcuru:
This is a minimal update to provide some attribution and proper naming of Apache Software Foundation projects and trademarks.  Thanks for considering.


zhzhan:
Thanks for the patch.

PR #13, opened 2016-06-21
In and Not IN filters are now simplified as set difference.

shubhamchopra:


zhzhan:
Thanks for the patch.

PR #22, opened 2016-08-10
Performance improvements and bug fixes with In/Not In filters

nickdevp:
When there are a large number of items for In/Not In clause, I am getting stack overflow errors as part of HBaseFilter.buildFilter. I've created a pull request that has the test cases + fixes.


nickdevp:
Additionally, the 'In' filter is currently doing full table scans as illustrated by the test case in commit '	b1f6f6c'. Please run this test without the fix '48cffbe' commit which clearly shows the filter is not being applied within the HbaseTableScanRDD, but instead being handled by spark after all rows are returned.

Sherry302:
Thanks for the PR.

PR #28, opened 2016-09-09
Make SHC work well with Spark2.0.0

Sherry302:


Sherry302:
Changes: 
1. Add an internal Logging instead of Spark Logging `org.apache.spark.Logging`, which has gone in Spark 2.0.0.
2. Use `CatalystSqlParser.parseDataType` instead of` DataTypeParser`, which has gone in Spark 2.0.0.
3. Throw an exception instead of using `SparkSqlSerializer`, which has gone in Spark 2.0.0.
4. Delete unused imports and an unit test.

Test:
All unit tests passed.

Sherry302:
Hi, @dongjoon-hyun Could you please review this? Thanks.

dongjoon-hyun:
Yep, @Sherry302 !

dongjoon-hyun:
BTW, we don't have CI so far? What about starting Travis CI for some suite?

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseFilter.scala
```diff
@@ -22,16 +22,12 @@ import java.util
 import org.apache.hadoop.hbase.filter.CompareFilter.CompareOp
 import org.apache.hadoop.hbase.filter.{Filter => HFilter, FilterList => HFilterList, _}
 import org.apache.hadoop.hbase.util.Bytes
-import org.apache.spark.Logging
```
dongjoon-hyun: FYI, it goes to `org.apache.spark.internal.Logging`, but I think your approach is right since we need to change package prefix someday.

pom.xml
```diff
@@ -32,17 +32,9 @@
...
-  <distributionManagement>
-    <repository>
-      <id>${repoid}</id>
-      <name>${reponame}</name>
-      <url>${repourl}</url>
-    </repository>
-  </distributionManagement> 
 
   <properties>
-    <spark.version>1.6.0</spark.version>
```
dongjoon-hyun: IMO, we had better change our version since our users need to use both libraries for their applications. What about changing the following line to make 2.0.0?

https://github.com/Sherry302/shc/blob/53ce222d9e95e7112c56f7dfc01caf10dc1b71a6/pom.xml#L31

pom.xml
```diff
@@ -32,17 +32,9 @@
   <packaging>jar</packaging>
   <name>HBase Spark Connector</name>
   <url>http://spark.apache.org/</url>
```
dongjoon-hyun: `https://github.com/hortonworks-spark/shc` ?

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/Logging.scala
```diff
@@ -0,0 +1,123 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.execution.datasources.hbase
+
+import org.apache.log4j.{Level, LogManager, PropertyConfigurator}
+import org.slf4j.{Logger, LoggerFactory}
+import org.slf4j.impl.StaticLoggerBinder
+
+trait Logging {
```
dongjoon-hyun: It seems to be a full copy of it of Spark?

src/test/scala/org/apache/spark/sql/DefaultSourceSuite.scala
```diff
@@ -136,17 +135,7 @@ class DefaultSourceSuite extends SHC with Logging {
     s.show
     assert(s.count() == df.count())
   }
-
-  test("IN filter, RDD") {
```
dongjoon-hyun: What about `ignore` instead of removing?

dongjoon-hyun:
For the scala-2.10, can we use `2.10.5` instead of `2.10.4` in the following line?
https://github.com/Sherry302/shc/blob/53ce222d9e95e7112c56f7dfc01caf10dc1b71a6/pom.xml#L302

dongjoon-hyun:
If this PR is going to branch `2.0.0` only, could you update README.md, too?

dongjoon-hyun:
I guess this will not work with Spark 1.6.1, but `README.md` describes about Spark 1.6.1.

dongjoon-hyun:
NIP. We had better upgrade examples for Spark 2.x. There are some deprecation warnings.

dongjoon-hyun:
Hi, @Sherry302 . I finished my first round, also confirmed that all tests are passed.
Could you fix the PR?

pom.xml
Sherry302: You are right. Currently SHC version is different from the branch name, which is corresponding to Spark version. It's a little confusing. SHC functionality of branch 1.6.1 and 2.0.0 are almost the same, so their versions  are the same - 1.0.0. The version of master branch is 1.1.0-SNAPSHOT.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/Logging.scala
Sherry302: Almost, but some unused code were deleted.

Sherry302:
Thanks for the review, @dongjoon-hyun. Deleting the deprecation warnings will go into next PR. I will update the PR soon.

pom.xml
dongjoon-hyun: Ur, I see what you mean now here. 
> their versions are the same - 1.0.0

Sherry302:
Hi, @dongjoon-hyun I have updated the PR. From now on, branch-1.6 (SHC version: 1.0.0) will be built against Spark 1.6.*; branch-2.0 (The name of branch 2.0.0 will be changed after this PR is merged. SHC version: 2.0.0) will be built against Spark 2.0.*; master will be built against Spark master. Besides using `Branches`, we may use `Tags` to maintain SHC in the future.

dongjoon-hyun:
Thank you, @Sherry302 . LGTM. It also passes all tests.

dongjoon-hyun:
BTW, after you merge this PR, what about renaming the branch from `2.0.0` into `branch-2.0`?

Sherry302:
Hi, @dongjoon-hyun Thanks for the review. Yes, I will rename 2.0.0 into branch-2.0.

Sherry302:
@dongjoon-hyun I have renamed 2.0.0 to branch-2.0. Thanks a lot.

dongjoon-hyun:
:)

PR #34, opened 2016-09-12
Avoid multiple copies of binary data during the conversion from Result to Row

Sherry302:


Sherry302:
Hi, @dongjoon-hyun Could you help review this? Thanks.

dongjoon-hyun:
LGTM!

nickdevp:
Just a quick question - why not just update Utils.hbaseFieldToScalaType directly?

Sherry302:
@nickdevp `Utils.hbaseFieldToScalaType` is also invoked in [`parseRowKey`](https://github.com/hortonworks-spark/shc/blob/master/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala#L90
https://github.com/hortonworks-spark/shc/blob/master/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala#L90), where `row` is used multiple times and can not be used as the return value of 'Utils.hbaseFieldToScalaType' directly.

Sherry302:
Hi, @nickdevp Could you please give your feedback about this PR? If this PR looks good for you, I will merge it.

nickdevp:
The code changes do fix the issue as I've described it. However, from a design point of view, you now have at least 2 places which need to be changed if the logic for converting a spark binary field to a scala type needs to be altered. 
As an alternate solution, I would probably add an argument to Utils.hbaseFieldToScalaField that specifies whether the data needs to be copied.

bikassaha:
@nickdevp I think for the current PR we chose to follow conventions for similar stuff in Spark. Opening a follow up issue to address the maintainability issue you raised. Its a valid concern.

PR #40, opened 2016-09-15
Add a cache for HBase connection object

weiqingy:


Sherry302:
@bikassaha @enis Could you please review this? I add a cache for HBase connection object. Thanks.

Sherry302:
Unit tests passed. Also, I have ran some SHC examples in HDP cluster, which shows the cache added works.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/Utils.scala
```diff
@@ -97,4 +101,30 @@ object Utils {
...
+    *
+    * From the Guava Docs: A Cache is similar to ConcurrentMap, but not quite the same. The most
+    * fundamental difference is that a ConcurrentMap persists all elements that are added to it until
+    * they are explicitly removed. A Cache on the other hand is generally configured to evict entries
+    * automatically, in order to constrain its memory footprint.  Note that this cache does not use
+    * weak keys/values and thus does not respond to memory pressure.
+    */
+  val removalListener = new RemovalListener[Configuration, Connection] {
+    override def onRemoval(rm: RemovalNotification[Configuration, Connection]): Unit = {
+      rm.getValue.close()
```
tedyu: Catch potential IOException

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/Utils.scala
```diff
@@ -97,4 +101,30 @@ object Utils {
...
+    }
+  }
+
+  val cache = CacheBuilder.newBuilder()
+    .maximumSize(100)
```
tedyu: Make this configurable ?

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/Utils.scala
```diff
@@ -113,18 +116,22 @@ object Utils {
     */
   val removalListener = new RemovalListener[Configuration, Connection] {
     override def onRemoval(rm: RemovalNotification[Configuration, Connection]): Unit = {
-      rm.getValue.close()
+      try{
+        rm.getValue.close()
+      } catch {
+        case e: IOException => throw e
```
tedyu: Who would be handling the IOE ?
Can we just log it ?

Sherry302: No one will handle it. SHC will stop running when this IOException happens. I think this exception is serious so I throw it out not log it. What do you think this?

tedyu: Since the connection is removed from cache - no one would be using it.

I think logging at warning level should be good.

Sherry302: OK. I have update the PR to log IOException.

tedyu:
lgtm

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/Utils.scala
```diff
@@ -97,4 +104,33 @@ object Utils {
...
+    override def onRemoval(rm: RemovalNotification[Configuration, Connection]): Unit = {
+      try{
+        rm.getValue.close()
+      } catch {
+        case e: IOException => logWarning("Fail to close HBase connection")
+      }
+    }
+  }
+
+  val cache = CacheBuilder.newBuilder()
```
enis: Who closes the Connection after we are done? Only closing is through cache removal, which would mean that these connections will be left open after the hbase job is done, no?

enis: We also should be careful for not re-using the same Connection across different users. You cannot give a Connection created for a user to another user.

Sherry302: Hi, @enis I just updated the PR [74f803a](https://github.com/hortonworks-spark/shc/pull/40/commits/74f803a7603d7f405b5f42847d80516f640b1edf).
1. Yes, closing is through cache removal. When SHC/Spark applications are going to exit, it will ask the cache to invalidate/close all of the connections (via invoking ShutdownHookManager.addShutdownHook). So SHC guarantees every connection is closed before the application exits.
2. Spark does not support multi-tenants environment and SHC is a Spark library, so SHC is supposed not to have the issue of re-using the same Connection across different users.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/Utils.scala
```diff
@@ -126,6 +128,7 @@ object Utils extends Logging{
 
   val cache = CacheBuilder.newBuilder()
     .maximumSize(SparkHBaseConf.connectCacheMaxSize)
+    .expireAfterAccess(SparkHBaseConf.cachedTime, TimeUnit.MINUTES)
```
enis: While the query is running, if the cache expires, then the connection will get closed and all scans from that connection will fail afterwards. This means that all queries have to be < cachedTime, no?

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/Utils.scala
```diff
@@ -97,4 +105,35 @@ object Utils {
   def toUTF8String(input: HBaseType, offset: Int, length: Int): UTF8String = {
     UTF8String.fromBytes(input.slice(offset, offset + length))
   }
+
+  /**
+    * A cache of Spark-HBase connections. Key is the reference of Configuration object as Configuration class
+    * does not implement hashCode() function.
+    *
```
shubhamchopra: I think the absence of hashCode() and equals() from Configuration will cause every dataframe targeting an HBase cluster have its own connection object in the current implementation. Ideally, we would want a single connection object be shared between creation of multiple dataframes as long as they are being created from the same Hbase cluster. So if we create dataframes from 2 tables from the same HBase clusters, we should be able to re-use the connection object. I don't think this patch handles it.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/Utils.scala
```diff
@@ -97,4 +105,35 @@ object Utils {
...
+  }
+
+  val cache = CacheBuilder.newBuilder()
+    .maximumSize(SparkHBaseConf.connectCacheMaxSize)
+    .expireAfterAccess(SparkHBaseConf.cachedTime, TimeUnit.MINUTES)
+    .removalListener(removalListener)
+    .build(
+      new CacheLoader[Configuration, Connection]() {
+        override def load(hbaseConf: Configuration): Connection = {
+          ConnectionFactory.createConnection(hbaseConf)
```
shubhamchopra: Can you please add some logging here, so we know new connections are being created.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/Utils.scala
```diff
@@ -97,4 +105,35 @@ object Utils {
...
+    * From the Guava Docs: A Cache is similar to ConcurrentMap, but not quite the same. The most
+    * fundamental difference is that a ConcurrentMap persists all elements that are added to it until
+    * they are explicitly removed. A Cache on the other hand is generally configured to evict entries
+    * automatically, in order to constrain its memory footprint.  Note that this cache does not use
+    * weak keys/values and thus does not respond to memory pressure.
+    */
+  val removalListener = new RemovalListener[Configuration, Connection] {
+    override def onRemoval(rm: RemovalNotification[Configuration, Connection]): Unit = {
+      try{
+        rm.getValue.close()
```
shubhamchopra: Some logging here as well, so we know when connections are closed.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/Utils.scala
Sherry302: Hi, @shubhamchopra @nickdevp Thanks for the comments.
Yes, the current cache implementation will create deferent HBase Connection objects if the parameter of Configuration objects are different (`ConnectionFactory.createConnection(hbaseConf)`). However, the tasks in the same executor will share the same Configuration object. The same to the driver in some procedures. So the current cache has already improved the performance a lot. Of course, with implementing `hashCode`() function and `equals`() of Configuration class, it will further reduce the number of creating Connection objects, but for that, we need to find a good way to override the `hashcode`() and `equal`() function. Need more time to look into HBase code, that’s why I did not implement it in the current PR. Will update the PR soon. Yes, will add test case and logs in the new PR.

Sherry302:
The PR has been updated.
The changes:
1. use ConcurrentHashMap instead of Guava cache.
2. use HBaseConnectionKey as the key of ConcurrentHashMap, which is the same as HConnectionKey in HBase.
3. Connection objects created in the ConcurrentHashMap will be closed by ShutdownHookManager before the Spark application exits.

Sherry302:
Hi, @tedyu @enis Could you please review this PR again? Thanks.

tedyu:
lgtm

PR #45, opened 2016-09-30
V2 Implementation of Connection Sharing

weiqingy:


weiqingy:
Changes:
Add a background thread to close the idle connections.

How to test:
1. add unit tests
2. test some examples in HDP cluster

weiqingy:
Hi, @tedyu @enis Could you please review this PR? Thanks.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionManager.scala
```diff
@@ -0,0 +1,218 @@
...
+  private final val DEFAULT_TIME_OUT: Long = SparkHBaseConf.connectionCloseDelay
+  private var timeout = DEFAULT_TIME_OUT
+
+  val housekeepingThread = new Thread(new Runnable {
+    override def run() {
+      while (true) {
+        try {
+          Thread.sleep(timeout)
+        } catch {
+          case e: InterruptedException =>
```
tedyu: Add a log

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionManager.scala
```diff
@@ -0,0 +1,218 @@
...
+    }
+  }
+
+  def performHousekeeping(forceClean: Boolean) = {
+    val tsNow: Long = System.currentTimeMillis()
+    connectionMap.synchronized {
+      connectionMap.foreach {
+        x => {
+          if(forceClean || ((x._2.rc <= 0) && (tsNow - x._2.timestamp > timeout))) {
+            x._2.c.close()
```
tedyu: There may be IOException coming out of close()

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionManager.scala
```diff
@@ -0,0 +1,218 @@
...
+        "if it is aborted by JVM while it's performing its task.")
+    }
+  }
+
+  def performHousekeeping(forceClean: Boolean) = {
+    val tsNow: Long = System.currentTimeMillis()
+    connectionMap.synchronized {
+      connectionMap.foreach {
+        x => {
+          if(forceClean || ((x._2.rc <= 0) && (tsNow - x._2.timestamp > timeout))) {
```
tedyu: Rename rc as refCount
rc < 0 indicates bug. Please handle separately

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionManager.scala
```diff
@@ -0,0 +1,218 @@
...
+import scala.collection.mutable
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.hbase.{TableName, HConstants}
+import org.apache.hadoop.hbase.client._
+import org.apache.hadoop.hbase.ipc.RpcControllerFactory
+import org.apache.hadoop.hbase.security.{User, UserProvider}
+import org.apache.spark.Logging
+
+private[spark] object HBaseConnectionManager extends Logging {
```
tedyu: Better rename this object - people would ask why not use HBaseConnectionManager in core

weiqingy:
Hi, @tedyu Thanks for the comments. I have updated the PR.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionManager.scala
weiqingy: Done.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionManager.scala
weiqingy: Done.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionManager.scala
weiqingy: Done.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionManager.scala
weiqingy: Will catch this. Done.

tedyu:
lgtm

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
```diff
@@ -0,0 +1,223 @@
...
+  private final val DEFAULT_TIME_OUT: Long = SparkHBaseConf.connectionCloseDelay
+  private var timeout = DEFAULT_TIME_OUT
+
+  val housekeepingThread = new Thread(new Runnable {
+    override def run() {
+      while (true) {
+        try {
+          Thread.sleep(timeout)
+        } catch {
+          case e: InterruptedException =>
```
tedyu: Restore interrupt status.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
```diff
@@ -0,0 +1,223 @@
...
+            logWarning(s"Someone tried to interrupt Housekeeping thread: ${e.getMessage}")
```
tedyu: Please use passive tense: housekeeping interrupted

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
weiqingy: If Thread.sleep() throws out an InterruptedException, it also clears the interrupt status. So when it gets an interruption, it just do some work and go to sleep again.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
weiqingy: Done.

weiqingy:
Thanks, @tedyu . I have updated the PR.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
```diff
@@ -0,0 +1,223 @@
...
+            logWarning(s"Housekeeping interrupted: ${e.getMessage}")
+        }
+        performHousekeeping(false)
+      }
+    }
+  })
+  housekeepingThread.setDaemon(true)
+  housekeepingThread.start()
+
+  def finalHouseKeeping() = {
```
enis: Instead, you probably want to name this close(), and implement the same semantics. The thread should be interrrupted, and closed here as well in the close method.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
```diff
@@ -0,0 +1,223 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  // For testing purpose only
+  def getConnection(key: HBaseConnectionKey, f: HBaseConnectionKey => Connection): SmartConnection =
+    connectionMap.synchronized {
+      val sc = connectionMap.getOrElseUpdate(key, new SmartConnection(f(key)))
```
enis: Is getOfElseUpdate() evaluate the second argument lazily?

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseResources.scala
```diff
@@ -86,15 +86,11 @@ trait ReferencedResource {
...
+  var connection: SmartConnection = _
   var rl: RegionLocator = _
 
   override def init(): Unit = {
-    import Utils.FuncConverter._
-
-    connection = Utils.connectionMap.computeIfAbsent(HBaseConnectionKey(relation.hbaseConf),
-      (k: HBaseConnectionKey) => Utils.getConnection(k))
-
+    connection = HBaseConnectionCache.getConnection(new HBaseConnectionKey(relation.hbaseConf))
```
enis: You can hide the HBaseConnectionKey from the rest of the code, and only use it internally inside the HBaseConnectionCache. Change the method signature to be getConnection(Configuration)

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
weiqingy: @enis Yeah that's a good point. This close() method (finalHouseKeeping()) is expected to be called when the process exits, so all the connections can be closed. When I implemented the first version I tended to interrupt the housekeeping thread when JVM exits and make the housekeeping thread do the final cleanup, however I found that when JVM decides to terminate, daemon threads are interrupted by JVM and even if you catch the InterruptedException, code in the catch clause (and any following code) may not get a chance to execute. That makes sense as when JVM decides to exit there are only daemon threads and it looks all the daemon threads are going to suffer a "sudden death". That's why I have to do all the cleanup in the shutdown hook. Please correct me if I've missed your point or if you have any suggestions. Sure I'll rename the function.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
weiqingy: Should be. I am writing some test code to double check, and I'll update the comment soon.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseResources.scala
weiqingy: You are right. My fault. Yes I am going to fix it.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
enis: It is fine if the JVM exit interrupts as well. The close() can be used independently though, not connected to the JVM shutdown handler. That is why it is good practice to to make sure that close() closes / interrupts the threads that we own.

weiqingy: You are right. I'll follow your comment.

weiqingy: @enis Just want to confirm that if client is allowed to call close(), we need to recreate the housekeeping thread when the client tries to get connections again?

enis: Well, in this patch you have the HBaseConnectionCache as a scala object (singleton) which means that once closed, it should not be used again. So in that sense, it should only be really closed when we are sure that there are no more users. We could have done it so that the ConnectionCache is an instantiatible class, but it is not worth doing it here I think. For Spark's purposes this should be fine.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
weiqingy: I have tested this. `getOfElseUpdate()` evaluates the second argument lazily.

src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseResources.scala
weiqingy: Done.

weiqingy:
Hi, @enis I have updated this PR based on your comments.

weiqingy:
@tedyu @enis Thanks for the review.

PR #52, opened 2016-10-24
Split src into `core` and `example`

weiqingy:
This PR is to split src into `core` and `example`.

The current repo structure of SHC is:
screen shot 2016-10-24 at 3 23 58 pm

With this PR, the repo structure of SHC will be:      
screen shot 2016-10-24 at 3 22 30 pm


weiqingy:
Hi, @dongjoon-hyun  Could you please help to review this PR? Thanks.

dongjoon-hyun:
It looks like testsuite are changed together. If then, could you describe more int PR description?

weiqingy:
Thanks for the review, @dongjoon-hyun.

PR #55, opened 2016-11-02
Minor change, improves performance of OR drastically.

shubhamchopra:
On local tests, using this to "OR" 2000 equals checks took around 600ms compared to 170s taken by the original version.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/ScanRange.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/ScanRange.scala
@@ -19,11 +19,12 @@ package org.apache.spark.sql.execution.datasources.hbase
 
 import scala.collection.mutable.ArrayBuffer
 import scala.math.Ordering
-
 import org.apache.hadoop.hbase.util.Bytes
 import org.apache.spark.Logging
 import org.apache.spark.sql.execution.datasources.hbase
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/ScanRange.scala
```diff
@@ -250,11 +251,15 @@ object ScanRange {
     }
   }
 
-  def or[T](
+  @tailrec def or[T](
```
weiqingy: Do we need to use "@tailrec" ? `or(right, left)`/ `or[T]` is called only twice in total.

weiqingy:
@shubhamchopra Thanks for this. Only two minor comments.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/ScanRange.scala
shubhamchopra: Just minimizing any additional stack frames being added due to this. This is a compiler directive, and shouldn't hurt performance. It is a tail call here, the compiler should pick up on it either ways, so this is just "trust but verify" approach.

weiqingy:
LGTM. Merging to master/branch-1.6/branch-2.0. Thanks.

PR #58, opened 2016-11-17
BUG-69053

mridulm:
a) Dont use case class for mutable state.
b) Remove broadcast when not required.
c) Fix MT issues in connection cache.
d) Fix encapsulation for some of the classes.

+CC @weiqingy, please review, thanks

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
@@ -80,13 +80,13 @@ private[spark] object HBaseConnectionCache extends Logging {
     }
   }
 
-  // for testing.
-  private[hbase] def resetCache(): Unit = {
+  // resetStats == true for testing, otherwise, it is not modified.
+  private[hbase] def resetCache(resetStats: Boolean = false): Unit = {
     connectionMap.synchronized {
       if (closed.get()) return
       connectionMap.values.foreach(conn => IOUtils.closeQuietly(conn) )
       connectionMap.clear()
-      cacheStat.reset()
+      if (resetStats) cacheStat.reset()
     }
   }
 
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
@@ -17,11 +17,14 @@
 
 package org.apache.spark.sql.execution.datasources.hbase
 
-import java.io.IOException
-import scala.collection.mutable
+import java.io.{Closeable, IOException}
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicLong}
 
+import org.apache.commons.io.IOUtils
+
+import scala.collection.mutable
 import org.apache.hadoop.conf.Configuration
-import org.apache.hadoop.hbase.{TableName, HConstants}
+import org.apache.hadoop.hbase.{HConstants, TableName}
 import org.apache.hadoop.hbase.client._
 import org.apache.hadoop.hbase.ipc.RpcControllerFactory
 import org.apache.hadoop.hbase.security.{User, UserProvider}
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
```diff
@@ -30,73 +33,91 @@ import org.apache.spark.Logging
...
-    try {
-      connectionMap.synchronized {
-        if(closed)
-          return
-        closed = true
+    connectionMap.synchronized {
+      if (closed.get()) return
+      try {
         housekeepingThread.interrupt()
+        resetCache()
```
mridulm: This change means that after close(), getStat will return (0, 0, 0)
If that is not the expectation, I can change it to only close the connections and leave the stats alone.

tedyu: Please leave the stats alone.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
```diff
@@ -30,73 +33,91 @@ import org.apache.spark.Logging
...
+  // 'duration' specified
+  private[hbase] def sleep(duration: Long, allowInterrupt: Boolean = false, allowClosed: Boolean = false): Unit = {
+    val startTime = System.currentTimeMillis()
+    var remaining = duration
+    while (remaining > 0) {
+      try {
+        Thread.sleep(remaining)
+      } catch {
+        case ex: InterruptedException if allowInterrupt => return
+        case ex: Exception => // ignore
```
tedyu: Why ignore the exception ?

weiqingy:
Hi, @mridulm  a, b and d, LGTM. For "c", could you elaborate what MT issues you are fixing? The reason that current test cases (`testBasic`, `testWithPressureWithoutClose`, `testWithPressureWithClose`) do not use `HBaseConnectionCache.connectionMap.synchronized` is there will be only one thread when doing `assert`. Of course, it does not hurt to add `HBaseConnectionCache.connectionMap.synchronized`. I'm neutral.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
mridulm: Exception is expected when there is an explicit interrupt.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
mridulm: Will do

mridulm:
@weiqingy The issues with (c) is not with the test suite specifically - though it exists there too.
Any shared variable which is accessed/modified from different threads needs to be protected : either synchronized or via locks (Atomic*) - timeout, closed, connectionMap : otherwise, updates need not necessarily be visible across threads.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseConnectionCache.scala
mridulm: resetStats == false by default now, preserving the stats even after a close()

bikassaha:
Should some/all of this go into branch-1 and brach-2.1

mridulm:
@bikassaha Ah, did not know we were maintaining different versions.
Once this goes through, will raise PR's against the other branches too. Thanks for letting me know.

PS: weirdly, dont get notifications for these comments ... strange.

weiqingy:
@mridulm Currently for SHC, except the master branch, we do not require people to raise PRs against other branches. If the PRs are important for other branches, we will do the backport. Of course, we do welcome people to raise PRs against all the branches. :) Thanks.

mridulm:
Great, thanks for clarifying @weiqingy !

PR #60, opened 2016-12-02
Preserve the columns order of the schema.

lins05:


weiqingy:
Hi, @lins05 Thanks for the PR. Sorry for the late response. Is there any error happened if the columns order of the schema is not preserved?

lins05:
@weiqingy no errors, just I was a bit surprised when I call `df.show()` and see the order of columns in the printed output is not the same as I provided in the catalog.

weiqingy:
I see. This PR does not change any behavior but might affect the performance (HashMap VS JTreeMap) . cc @dongjoon-hyun

weiqingy:
@lins05  TreeMap uses dictionary order instead of input order. For example, given ('b' -> 2, 'a' -> 1, 'c' ->3), TreeMap seems to return ('a' -> 1, 'b' -> 2, 'c' -> 3) in a dictionary order, not the input order. In the test case, if the input order of catalog is not ("col00", "col01", ..., "col8"), the test case will fail.

lins05:
Ah, right, I should use `LinkedHashMap`.

dongjoon-hyun:
Hi, @lins05 , any update?

lins05:
@dongjoon-hyun @weiqingy I have updated the code and the tests, please take a look.

weiqingy:
LGTM.

weiqingy:
@lins05 Could you make a PR to master and branch 2.1? Thanks.

lins05:
@weiqingy 

- PR to branch-2.1: https://github.com/hortonworks-spark/shc/pull/70
- PR to master: https://github.com/hortonworks-spark/shc/pull/71

Issue #64, opened 2016-12-14
Support of Array data type


NikolayTsvetkov:
Hello,
We store arrays in Hbase serialized with Avro schema and we need to get the data deserialized by the connector. We have tried using the Avro feature that worked perfectly for other complex types, but in this case we serialize the data only using a simple array schema {"type": "array", "items": ["long","null"]} and not the complete record schema.
The connector fails with ClassCastException while trying to deserialize and cast to GenericRecord.
Do you have any plans of supporting such schemas or array types out of the box ?
Thank you in advance!

Cheers,
Nikolay

PR #65, opened 2016-12-16
Update SchemaConverters.scala

NikolayTsvetkov:
If the reader returns Object instead of GenericRecord the **deserialize** method will support all the Avro schemas. It currently casts to GenericRecord and it doesn't work for simple types and union schemas.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SchemaConverters.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SchemaConverters.scala
@@ -401,11 +401,11 @@ object AvroSedes {
     }
   }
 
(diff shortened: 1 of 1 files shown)


weiqingy:
Hi, @NikolayTsvetkov Thanks for this PR. Will merge this to Master. I think this PR also resolves the [Issue#64](https://github.com/hortonworks-spark/shc/issues/64), right?

NikolayTsvetkov:
Hi @weiqingy,
Yes, in fact I've modified the code in order to solve Issue#64.
It would be great if it goes merged to branches 2.0 and 2.1 as well.
I will add some additional examples and mention this feature in the read me as well.

Cheers,
Nikolay

weiqingy:
Thanks. Merged to Master/branch-2.0/branch-2.1. Yes, it would be great to add some additional examples and update readme.

PR #77, opened 2017-01-11
Fix the issue that different columns cannot be written into the same column family

weiqingy:
This PR is to fix Issue#62 and Issue#44. The solution is [scrapinghub@d73da66](https://github.com/scrapinghub/shc/commit/d73da66)
Thanks, @lins05 . All credit should go to @lins05.


weiqingy:
Merge this PR to branch-1.6/branch-2.0/branch-2.1/master.

PR #78, opened 2017-01-12
Explict Array[Byte] type for the HRF inside the HBaseFilter

merlintang:
The type mismatch would bother users in the Intelj. @dongjoon-hyun @weiqingy please review it.


dongjoon-hyun:
How do you think of this @weiqingy ?

weiqingy:
@dongjoon-hyun We can make this PR more complete. I just searched the SHC code in Intelj, there are four places in [object BoundRange](https://github.com/hortonworks-spark/shc/blob/master/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/ScanRange.scala#L419) have the issue. Could you please fix all of them in this PR? Sorry for the late reply. @merlintang

weiqingy:
Merged to master. Thanks.

PR #86, opened 2017-01-30
Make avro definition in catalog backwards compatible

weiqingy:
Make avro definition in catalog backwards compatible: use `"avro": schema` instead

## How was this patch tested?
Pass current unit tests.

### DIFF
--- a/examples/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/AvroRecord.scala
+++ b/examples/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/AvroRecord.scala
@@ -35,14 +35,14 @@ object AvroRecord {
       val p = new Schema.Parser
       p.parse(schemaString)
     }
-    val user1 = new GenericData.Record(avroSchema);
-    user1.put("name", "Alyssa");
-    user1.put("favorite_number", 256);
+    val user1 = new GenericData.Record(avroSchema)
+    user1.put("name", "Alyssa")
+    user1.put("favorite_number", 256)
 
-    val user2 = new GenericData.Record(avroSchema);
-    user2.put("name", "Ben");
-    user2.put("favorite_number", 7);
-    user2.put("favorite_color", "red");
+    val user2 = new GenericData.Record(avroSchema)
+    user2.put("name", "Ben")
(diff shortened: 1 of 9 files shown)


weiqingy:
We want to use “Phoenix” as the default coder, so the change will still break the existing SHC based users who have used PrimitiveType as coder, such as [this case](https://github.com/hortonworks-spark/shc/blob/master/core/src/test/scala/org/apache/spark/sql/DataTypeSuite.scala#L51). User need to add `"tableCoder":"PrimitiveType"` to tell SHC they want to use “PrimitiveType” instead of “Phoenix”.

@dongjoon-hyun @bikassaha Could you please review this PR? Thanks. cc: @merlintang

examples/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/AvroRecord.scala
```diff
@@ -35,14 +35,14 @@ object AvroRecord {
       val p = new Schema.Parser
       p.parse(schemaString)
     }
-    val user1 = new GenericData.Record(avroSchema);
-    user1.put("name", "Alyssa");
-    user1.put("favorite_number", 256);
```
dongjoon-hyun: Are there more example like this?

dongjoon-hyun: If then, could we clean up them in a single PR?

dongjoon-hyun:
+1, LGTM for the changes.

examples/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/AvroRecord.scala
weiqingy: Yes, there is another one in [`AvroSource.scala`](https://github.com/hortonworks-spark/shc/blob/master/examples/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/AvroSource.scala#L49). I just fixed this in [`PR#87`](https://github.com/hortonworks-spark/shc/pull/87).

weiqingy:

PR #87, opened 2017-01-31
Make 'buildSchema' happen once for task instead of per row

weiqingy:
This PR is to fix review comments in PR#84.
In the `decodeCompositeRowKey` function, make 'buildSchema' happen once for task instead of per row.

## How was this patch tested?
Pass current unit tests.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/types/Phoenix.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/types/Phoenix.scala
@@ -28,6 +28,7 @@ import org.apache.spark.sql.execution.datasources.hbase._
 import org.apache.spark.sql.types._
 
 class Phoenix(f:Option[Field] = None) extends SHCDataType {
+  private var schema: RowKeySchema = null
 
   def fromBytes(src: HBaseType): Any = {
     if (f.isDefined) {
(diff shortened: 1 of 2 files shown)


weiqingy:
Hi, @enis @dongjoon-hyun Could you please help to review it? Thanks.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/types/Phoenix.scala
```diff
@@ -60,15 +61,7 @@ class Phoenix(f:Option[Field] = None) extends SHCDataType {
...
-    def buildSchema(): RowKeySchema = {
-      val builder: RowKeySchemaBuilder = new RowKeySchemaBuilder(keyFields.length)
-      keyFields.foreach{ x =>
-        builder.addField(buildPDatum(x.dt), false, SortOrder.getDefault)
-      }
-      builder.build
-    }
-    val schema: RowKeySchema = buildSchema()
-
+    if (schema == null) schema = buildSchema(keyFields)
```
dongjoon-hyun: @weiqingy .
What if he existing `schema` is generated for the different arguments?
I mean different `keyFields` at `buildSchema(keyFields)`?

weiqingy: Hi, @dongjoon-hyun Thanks for the review. For one Phoenix object, the value of the parameter `keyFields`passed into `buildSchema` are the same. 

`decodeCompositeRowKey` is only called by `relation.catalog.shcTableCoder.decodeCompositeRowKey(r, relation.catalog.getRowKey)` in [`buildRow`](https://github.com/hortonworks-spark/shc/blob/master/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala#L91) function in `HBaseTableScanRDD`.  `shcTableCoder` is a [variable member of HBaseTableCatalog](https://github.com/hortonworks-spark/shc/blob/master/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala#L141) . If user choose Phoenix as the table coder in their applications,  `shcTableCoder` will be a Phoenix object. 

One catalog is mapping to one HBase table. In one Hbase table, for each row key, their metadata (aka. `keyFields`) are the same. For example, the row key of [the catalog](https://github.com/hortonworks-spark/shc/blob/master/core/src/test/scala/org/apache/spark/sql/CompositeKeySuite.scala#L58) is `key1:key2`, the `keyFields`(aka. metadata)of  row keys contain the information of the column family name, the column name, the type of [key1](https://github.com/hortonworks-spark/shc/blob/master/core/src/test/scala/org/apache/spark/sql/CompositeKeySuite.scala#L60) and [key2](https://github.com/hortonworks-spark/shc/blob/master/core/src/test/scala/org/apache/spark/sql/CompositeKeySuite.scala#L61).

dongjoon-hyun: In that case, I'm wondering if we can use `lazy val schema` instead of `var schema`?

weiqingy:

PR #95, opened 2017-02-17
Add CatalogVersion to fix the incompatibility issue

weiqingy:

1. Add `CatalogVersion` to fix the incompatibility issue
2. Add an example

## How was this patch tested?
Pass the current unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Apache Spark - Apache HBase Connector
 
-The [Apache Spark](https://spark.apache.org/) - [Apache HBase](https://hbase.apache.org/) Connector is a library to support Spark accessing HBase table as external data source or sink. With it, user can operate HBase with Spark-SQL on data frame level.
+The [Apache Spark](https://spark.apache.org/) - [Apache HBase](https://hbase.apache.org/) Connector is a library to support Spark accessing HBase table as external data source or sink. With it, user can operate HBase with Spark-SQL on data frame level. 
 
 With the data frame support, the lib leverages all the optimization techniques in catalyst, and achieves data locality, partition pruning, predicate pushdown, Scanning and BulkGet, etc.
 
--- a/README.md
+++ b/README.md
@@ -1,19 +1,16 @@
 # Apache Spark - Apache HBase Connector
 
-The [Apache Spark](https://spark.apache.org/) - [Apache HBase](https://hbase.apache.org/) Connector is a library to support Spark accessing HBase table as external data source or sink. With it, user can operate HBase with Spark-SQL on data frame level. 
+The [Apache Spark](https://spark.apache.org/) - [Apache HBase](https://hbase.apache.org/) Connector is a library to support Spark accessing HBase table as external data source or sink. With it, user can operate HBase with Spark-SQL on data frame level.
 
-With the data frame support, the lib leverages all the optimization techniques in catalyst, and achieves data locality, partition pruning, predicate pushdown, Scanning and BulkGet, etc. 
+With the data frame support, the lib leverages all the optimization techniques in catalyst, and achieves data locality, partition pruning, predicate pushdown, Scanning and BulkGet, etc.
 
 ## Catalog
 For each table, a catalog has to be provided,  which includes the row key, and the columns with data type with predefined column families, and defines the mapping between hbase column and table schema. The catalog is user defined json format.
 
 ## Datatype conversion
-Java primitive types is supported. In the future, other data types will be supported, which relies on user specified sedes. Take avro as an example. User defined sedes will be responsible to convert byte array to avro object, and connector will be responsible to convert avro object to catalyst supported datatypes. 
+Java primitive types is supported. In the future, other data types will be supported, which relies on user specified serdes. There are three internal serdes supported in SHC: Avro, Phoenix, PrimitiveType. User can specify which serde they want to use by defining 'tableCoder' in their catalog. For this, please refer to examples and unit tests. Take Avro as an example. User defined serdes will be responsible to convert byte array to Avro object, and connector will be responsible to convert Avro object to catalyst supported data types. When user define a new serde, they need to make it 'implement' the trait 'SHCDataType'.
 
 Note that if user want dataframe to only handle byte array, the binary type can be specified. Then user can get the catalyst row with each column as a byte array. User can further deserialize it with customized deserializer, or operate on the RDD of the data frame directly.
-
-## Solve the conflicts of Catalyst datatype order and HBase byte array order
-The libary automatically handle the orderness between the conflicts between the java data type (Short, Integer, Long, Float, Double, String, etc) and HBase bytearray order, which means as long as the format saved in HBase is consistent with the Java native data format, the libary will be able to take care of pruning and comparison instead of relying on a specific import tools.
 
 ## Data locality
 When the spark work node co-located with hbase region servers, data locality is achieved by identifying the region server location, and co-locate the executor with the region server. Each executor will only perform Scan/BulkGet on the part of the data that co-locates on the same host. 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Apache Spark - Apache HBase Connector
 
-The [Apache Spark](https://spark.apache.org/) - [Apache HBase](https://hbase.apache.org/) Connector is a library to support Spark accessing HBase table as external data source or sink. With it, user can operate HBase with Spark-SQL on data frame level. 
+The [Apache Spark](https://spark.apache.org/) - [Apache HBase](https://hbase.apache.org/) Connector is a library to support Spark accessing HBase table as external data source or sink. With it, user can operate HBase with Spark-SQL on data frame level.
 
 With the data frame support, the lib leverages all the optimization techniques in catalyst, and achieves data locality, partition pruning, predicate pushdown, Scanning and BulkGet, etc.
 
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
@@ -17,6 +17,8 @@
 
 package org.apache.spark.sql.execution.datasources.hbase
 
+import java.lang.NumberFormatException
+
 import scala.collection.mutable
 
 import org.json4s.JsonAST.JObject
@@ -32,9 +34,9 @@ import org.apache.spark.sql.execution.datasources.hbase.types._
 
 case class CatalogVersion(major: Int, minor: Int) extends Comparable[CatalogVersion] {
   override def compareTo(o: CatalogVersion): Int = {
-    if(major > o.major)
+    if (major > o.major)
       1
-    else if(major == o.major)
+    else if (major == o.major)
       minor - o.minor
     else
       -1
@@ -45,12 +47,23 @@ case class CatalogVersion(major: Int, minor: Int) extends Comparable[CatalogVers
 
 object CatalogVersion {
   def apply(s: String): CatalogVersion = {
-    if(!s.matches("^[0-9]*\\.[0-9]*$"))
+    // Valid versions: "1.3", "1"
+    // Invalid versions: ".3"
+    if (!s.matches("^[0-9]{1,9}(\\.[0-9]{1,9})?$"))
       throw new IllegalArgumentException("Invalid version: " + s)
 
     val arr: Array[String] = s.split("\\.")
-    val m = Integer.parseInt(arr(0))
-    val n = Integer.parseInt(arr(1))
+
+    var m: Int = 0
+    var n: Int = 0
+    try {
+      m = Integer.parseInt(arr(0)) // must always have a major version number
+      if (arr(1) != null && !arr(1).isEmpty) // minor version number is optional
+        n = Integer.parseInt(arr(1))
+    } catch {
+      case e: NumberFormatException =>
+        throw new IllegalArgumentException("Invalid version: " + s)
+    }
     CatalogVersion(m, n)
   }
 }
@@ -208,6 +221,8 @@ case class HBaseTableCatalog(
   validateCatalogDef()
 }
 
+class CatalogDefinitionException(msg: String) extends Exception(msg)
+
 object HBaseTableCatalog {
   val newTable = "newtable"
   // The json string specifying hbase catalog information
@@ -231,7 +246,7 @@ object HBaseTableCatalog {
   val length = "length"
   val fCoder = "coder"
   val tableCoder = "tableCoder"
-  // The version num of catalog
+  // The version number of catalog
   val cVersion = "version"
   /**
    * User provide table schema definition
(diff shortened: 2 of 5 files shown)


weiqingy:
Hi, @dongjoon-hyun @bikassaha Could you please help to review this PR? Thanks. cc: @merlintang

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
```diff
@@ -30,6 +30,31 @@ import org.apache.spark.sql.catalyst.util.DataTypeParser
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.execution.datasources.hbase.types._
 
+case class CatalogVersion(major: Int, minor: Int) extends Comparable[CatalogVersion] {
+  override def compareTo(o: CatalogVersion): Int = {
+    if(major > o.major)
```
dongjoon-hyun: nit. `if(` -> `if (`?

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
```diff
@@ -30,6 +30,31 @@ import org.apache.spark.sql.catalyst.util.DataTypeParser
...
+      1
+    else if(major == o.major)
```
dongjoon-hyun: here, too.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
```diff
@@ -30,6 +30,31 @@ import org.apache.spark.sql.catalyst.util.DataTypeParser
...
+    else
+      -1
+  }
+
+  override def toString: String = major + "." + minor
+}
+
+object CatalogVersion {
+  def apply(s: String): CatalogVersion = {
+    if(!s.matches("^[0-9]*\\.[0-9]*$"))
```
dongjoon-hyun: `if(` -> `if (`.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
```diff
@@ -30,6 +30,31 @@ import org.apache.spark.sql.catalyst.util.DataTypeParser
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.execution.datasources.hbase.types._
 
+case class CatalogVersion(major: Int, minor: Int) extends Comparable[CatalogVersion] {
+  override def compareTo(o: CatalogVersion): Int = {
```
dongjoon-hyun: What if `o` is null? Can we implement this as null-safe? Otherwise, NPE might occur.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
dongjoon-hyun: This seems to match `dot` string; `.`. Then, parse error might occurs on line 52~53.
Could we change `^[0-9]*\\.[0-9]*$` into `^[0-9]+\\.[0-9]+$`?

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
```diff
@@ -206,6 +231,8 @@ object HBaseTableCatalog {
   val length = "length"
   val fCoder = "coder"
   val tableCoder = "tableCoder"
+  // The version num of catalog
```
dongjoon-hyun: nit. `num` -> `number`.

core/src/test/scala/org/apache/spark/sql/PhoenixSuite.scala
```diff
@@ -61,7 +61,7 @@ object PRecord {
 
 class PhoenixSuite extends SHC with Logging {
   override def catalog = s"""{
-                    |"table":{"namespace":"default", "name":"table1", "tableCoder":"Phoenix"},
+                    |"table":{"namespace":"default", "name":"table1", "tableCoder":"Phoenix", "version":"2.0"},
```
dongjoon-hyun: I'm just wondering. Is this okay? It looks too great change.
Next time, we need `3.0`?

examples/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/DataCoder.scala
```diff
@@ -0,0 +1,109 @@
...
+ */
+
+package org.apache.spark.sql.execution.datasources.hbase.examples
+
+import org.apache.spark.sql.execution.datasources.hbase._
+import org.apache.spark.{SparkConf, SparkContext}
+import org.apache.spark.sql.{SQLContext, DataFrame}
+
+case class DCRecord(col00: String,
+                   col01: Int,
```
dongjoon-hyun: nit, alignment?

examples/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/DataCoder.scala
```diff
@@ -0,0 +1,109 @@
...
+                |"col4":{"cf":"CF2", "col":"COL4", "type":"int"},
+                |"col5":{"cf":"CF3", "col":"COL5", "type":"bigint"},
+                |"col6":{"cf":"CF3", "col":"COL6", "type":"smallint"},
+                |"col7":{"cf":"CF4", "col":"COL7", "type":"string"},
+                |"col8":{"cf":"CF4", "col":"COL8", "type":"tinyint"}
+                |}
+                |}""".stripMargin
+
+  def main(args: Array[String]){
+    val sparkConf = new SparkConf().setAppName("CompositeKeyTest")
```
dongjoon-hyun: Since this is inside `example`, can we use `CompositeKeyExample` here?

examples/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/DataCoder.scala
```diff
@@ -0,0 +1,109 @@
...
+
+    def withCatalog(cat: String): DataFrame = {
+      sqlContext
+        .read
+        .options(Map(HBaseTableCatalog.tableCatalog->cat))
+        .format("org.apache.spark.sql.execution.datasources.hbase")
+        .load()
+    }
+
+    //populate table with composite key
```
dongjoon-hyun: nit. `// populate`

examples/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/DataCoder.scala
```diff
@@ -0,0 +1,109 @@
...
+    val data = (0 to 255).map { i =>
+      DCRecord(i)
+    }
+
+    sc.parallelize(data).toDF.write.options(
+      Map(HBaseTableCatalog.tableCatalog -> cat, HBaseTableCatalog.newTable -> "5"))
```
dongjoon-hyun: indentation?

weiqingy:
Thanks for the quick feedback, @dongjoon-hyun . I will update the PR soon.

README.md
```diff
@@ -1,20 +1,17 @@
...
 
 Note that if user want dataframe to only handle byte array, the binary type can be specified. Then user can get the catalyst row with each column as a byte array. User can further deserialize it with customized deserializer, or operate on the RDD of the data frame directly.
 
-## Solve the conflicts of Catalyst datatype order and HBase byte array order
-The libary automatically handle the orderness between the conflicts between the java data type (Short, Integer, Long, Float, Double, String, etc) and HBase bytearray order, which means as long as the format saved in HBase is consistent with the Java native data format, the libary will be able to take care of pruning and comparison instead of relying on a specific import tools.
```
bikassaha: Why is this removed?

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
```diff
@@ -30,6 +30,31 @@ import org.apache.spark.sql.catalyst.util.DataTypeParser
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.execution.datasources.hbase.types._
 
+case class CatalogVersion(major: Int, minor: Int) extends Comparable[CatalogVersion] {
```
bikassaha: Major and minor version is probably an overkill.

core/src/test/scala/org/apache/spark/sql/PhoenixSuite.scala
bikassaha: Since we just added this, it looks weird to start with 2.0 instead of 1.0

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
```diff
@@ -220,8 +247,20 @@ object HBaseTableCatalog {
...
-    // If 'tableCoder' is not specified in the catalog, 'Phoenix' will be the table coder
-    val tCoder = tableMeta.getOrElse(tableCoder, SparkHBaseConf.Phoenix).asInstanceOf[String]
+
+    // The snippet below is to fix the incompatibility issue caused by the change of adding
+    // Phoenix as coder. Before the change, the catalog version is 1.0. If the catalog
+    // version specified by user is later than 1.0, tableCoder must be specified too.
+    // The default value of the catalog version is 1.0. When the catalog version is 1.0,
+    // the default value of tableCoder is PrimitiveType.
+    val vNum = tableMeta.getOrElse(cVersion, "1.0").asInstanceOf[String]
+    if (CatalogVersion(vNum).compareTo(CatalogVersion("1.0")) == 1){
```
bikassaha: We may be using version often if we keep making changes to the catalog def. So would be good to have an int representation that can be easily compared without having to do string manipulation. If we drop minor version then it may be straightforward to do.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
```diff
@@ -220,8 +247,20 @@ object HBaseTableCatalog {
...
+      if (tableMeta.get(tableCoder).isEmpty)
+        throw new UnsupportedClassVersionError("Please specify 'tableCoder' in your catalog " +
```
bikassaha: UnsupportedClassVersionError? These have specific meanings in Java. So its good to not use them outside of those well defined meanings. In this case we could use InvalidParameterException or something similar.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
```diff
@@ -220,8 +247,20 @@ object HBaseTableCatalog {
...
+          "if the catalog version is later than 1.0")
+    }
+
+    val tCoder = tableMeta.getOrElse(tableCoder, SparkHBaseConf.PrimitiveType).asInstanceOf[String]
```
bikassaha: IMO the safe way to write this kind of logic would be
if (version == 1.0) { tableCoder = "primitive" } else { tableCoder = tableMeta.getOrElseThrowException } 

The way its currently written tCoder = tableMeta.getOrElse() is outside the version logic but dependent on it. And in future, someone could change the version logic and leave this line unchanged leading to bugs.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
weiqingy: I understand using a single major version would be much simpler. It's just a little weird to me. Nowadays people have got used to version numbers and all versions are at least like "x.y", from iOS, Windows, and open source software. What's more, if "1" is a valid version number, user would probably expect "1.0" valid.

Still, I'd prefer to handle version in a more general way, leaving room for future possible minor version upgrades. Java 9 is going to have its own Version class and we probably just replace this version class with the Java one then.

core/src/test/scala/org/apache/spark/sql/PhoenixSuite.scala
weiqingy: This version is the one with "Phoenix" feature in. I suppose it should be 2.0 so the previous version (without Phoenix support) could be 1.0?

README.md
weiqingy: This is just for PrimitiveType. PrimitiveType invokes HBase APi `Bytes.toBytes` to encode data, so it needs to do extra work to handle the conflicts of byte array order between java data type and HBase bytearray order. Now SHC has added Phoenix as coder, which relies on Phoenix data types to encode/decode data.

core/src/test/scala/org/apache/spark/sql/PhoenixSuite.scala
weiqingy: This version 2.0 is the one with "Phoenix" feature in. I suppose the previous version (without Phoenix support) could be 1.0.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
weiqingy: I am actually neutral about this - I noticed in the Java library source code, they don't check whether the parameter of compareTo is null or not, like in Integer.compareTo, String.compareTo. They just choose to let NPE occur. I suppose maybe compareTo is a so widely used method and everybody knows a null object should not be passed, and if it does happen, a NPE suffices.

I am personally OK with either: NPE or IllegalArgumentException.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
weiqingy: Done.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
weiqingy: Done.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
weiqingy: I have updated the PR for this.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
weiqingy: I created a class `CatalogDefinitionException` for this.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
weiqingy: Thanks. I have updated the PR for this.

dongjoon-hyun:
Is it ready for review?

weiqingy:
@dongjoon-hyun Yes, it's ready.

dongjoon-hyun:
Could you resolve the conflicts?

weiqingy:
@dongjoon-hyun The conflicts have been fixed.

dongjoon-hyun:
LGTM

weiqingy:

PR #106, opened 2017-03-16
Return exception when HBaseTableCatalog.newTable <= 3

khampson:

Return an exception when the option HBaseTableCatalog.newTable <= 3. Previously, no exception would be returned, and a downstream failure would occur when the HBase API could not find a table with the specified name.

## How was this patch tested?

I manually tested the patch by specifiying write code like this:
`        sqlRes.write.options(
                Map(HBaseTableCatalog.tableCatalog -> catalogTestWrite, HBaseTableCatalog.newTable -> "3"))
        .format("org.apache.spark.sql.execution.datasources.hbase")
        .save()
`

and confirming that the exception was thrown, and then by changing it to "5" and confirming that the table was written.


weiqingy:
Thanks, @khampson . Merged to master/branch-2.1/branch-2.0/branch-1.6.

PR #107, opened 2017-03-23
[WIP] Support multiple secure HBase clusters

weiqingy:
## What changes were proposed in this pull request?
Support multiple secure HBase clusters:

- [x] add SHCCredentialProvider
1. Obtain token 
2. Reissue token
3. Work with cache

- [ ] others:

1. Get token from custom delegation manager for SHC
2. In the driver, serialize token for SHC tasks; In the executor, deserialize token and add to the current UGI

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

PR #109, opened 2017-03-28
Add a HBaseCredentialsManager

jerryshao:
CC @bikassaha @weiqingy .

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
@@ -23,9 +23,9 @@ import scala.collection.mutable
 import scala.language.existentials
 import scala.util.control.NonFatal
 
-import org.apache.hadoop.hbase.HBaseConfiguration
+import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.hbase.security.token.TokenUtil
-import org.apache.hadoop.security.Credentials
+import org.apache.hadoop.security.{Credentials, UserGroupInformation}
 import org.apache.hadoop.security.token.{Token, TokenIdentifier}
 
 import org.apache.spark.util.{ThreadUtils, Utils}
@@ -33,7 +33,7 @@ import org.apache.spark.util.{ThreadUtils, Utils}
 final class HBaseCredentialsManager private() extends Logging {
   private class TokenInfo(
       val expireTime: Long,
-      val conf: HBaseConfiguration,
+      val conf: Configuration,
       val token: Token[_ <: TokenIdentifier])
   private val tokensMap = new mutable.HashMap[String, TokenInfo]
 
@@ -54,9 +54,9 @@ final class HBaseCredentialsManager private() extends Logging {
   /**
    * Get HBase credential from specified cluster name.
    */
-  def getCredentialsForCluster(hbaseConf: HBaseConfiguration): Credentials = {
+  def getCredentialsForCluster(conf: Configuration): Credentials = {
     val credentials = new Credentials()
-    val identifier = clusterIdentifier(hbaseConf)
+    val identifier = clusterIdentifier(conf)
 
     val tokenOpt = this.synchronized {
       tokensMap.get(identifier)
@@ -67,7 +67,7 @@ final class HBaseCredentialsManager private() extends Logging {
       credentials.addToken(tokenOpt.get.token.getService, tokenOpt.get.token)
     } else {
       // Acquire a new token if not existed or old one is expired.
-      val tokenInfo = getNewToken(hbaseConf)
+      val tokenInfo = getNewToken(conf)
       this.synchronized {
         tokensMap.put(identifier, tokenInfo)
       }
@@ -77,6 +77,11 @@ final class HBaseCredentialsManager private() extends Logging {
     }
 
     credentials
+  }
+
+  def isCredentialsRequired(conf: Configuration): Boolean = {
+    UserGroupInformation.isSecurityEnabled &&
+      conf.get("hbase.security.authentication") == "kerberos"
   }
 
   private def isTokenExpired(expireTime: Long): Boolean = {
@@ -120,23 +125,23 @@ final class HBaseCredentialsManager private() extends Logging {
     }
   }
 
-  private def getNewToken(hbaseConf: HBaseConfiguration): TokenInfo = {
-    val token = TokenUtil.obtainToken(hbaseConf)
+  private def getNewToken(conf: Configuration): TokenInfo = {
+    val token = TokenUtil.obtainToken(conf)
     val tokenIdentifier = token.decodeIdentifier()
     val expireTime =
       expectedExpireTime(tokenIdentifier.getIssueDate, tokenIdentifier.getExpirationDate)
     logInfo(s"Obtain new token with expiration time $expireTime")
-    new TokenInfo(expireTime, hbaseConf, token)
+    new TokenInfo(expireTime, conf, token)
   }
 
-  private def clusterIdentifier(hbaseConf: HBaseConfiguration): String = {
-    require(hbaseConf.get("zookeeper.znode.parent") != null &&
-      hbaseConf.get("hbase.zookeeper.quorum") != null &&
-      hbaseConf.get("hbase.zookeeper.property.clientPort") != null)
+  private def clusterIdentifier(conf: Configuration): String = {
+    require(conf.get("zookeeper.znode.parent") != null &&
+      conf.get("hbase.zookeeper.quorum") != null &&
+      conf.get("hbase.zookeeper.property.clientPort") != null)
 
-    hbaseConf.get("zookeeper.znode.parent") + "-"
-      hbaseConf.get("hbase.zookeeper.quorum") + "-"
-      hbaseConf.get("hbase.zookeeper.property.clientPort")
+    conf.get("zookeeper.znode.parent") + "-"
+      conf.get("hbase.zookeeper.quorum") + "-"
+      conf.get("hbase.zookeeper.property.clientPort")
   }
 }
 
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
@@ -48,7 +48,8 @@ final class HBaseCredentialsManager private() extends Logging {
     override def run(): Unit = Utils.logUncaughtExceptions(updateTokensIfRequired())
   }
 
-  tokenUpdater.schedule(tokenUpdateRunnable, nextRefresh, TimeUnit.MILLISECONDS)
+  tokenUpdater.scheduleAtFixedRate(
+    tokenUpdateRunnable, nextRefresh, nextRefresh, TimeUnit.MILLISECONDS)
 
   /**
    * Get HBase credential from specified cluster name.
@@ -58,14 +59,15 @@ final class HBaseCredentialsManager private() extends Logging {
       hbaseConf: HBaseConfiguration): Credentials = synchronized {
     val credentials = new Credentials()
 
+    val tokenOpt = tokensMap.get(hbaseCluster)
     // If token is existed and not expired, directly return the Credentials with tokens added in.
-    if (tokensMap.contains(hbaseCluster) && !isTokenExpired(tokensMap(hbaseCluster).expireTime)) {
-      credentials.addToken(tokensMap(hbaseCluster).token.getService, tokensMap(hbaseCluster).token)
+    if (tokenOpt.isDefined && !isTokenExpired(tokenOpt.get.expireTime)) {
+      credentials.addToken(tokenOpt.get.token.getService, tokenOpt.get.token)
     } else {
       // Acquire a new token if not existed or old one is expired.
       val tokenInfo = getNewToken(hbaseConf)
       tokensMap.put(hbaseCluster, tokenInfo)
-      logInfo(s"Obtain new token from cluster $hbaseCluster")
+      logInfo(s"Obtain new token for cluster $hbaseCluster")
 
       credentials.addToken(tokenInfo.token.getService, tokenInfo.token)
     }
@@ -108,13 +110,9 @@ final class HBaseCredentialsManager private() extends Logging {
         }
       } catch {
         case NonFatal(e) =>
-          logWarning("Error while trying to fetch tokens from HBase cluster," +
-            " will try again in 1 minute", e)
+          logWarning("Error while trying to fetch tokens from HBase cluster", e)
       }
     }
-
-    logInfo(s"Scheduling token update from HBase clusters in $nextRefresh ms.")
-    tokenUpdater.schedule(tokenUpdateRunnable, nextRefresh, TimeUnit.MILLISECONDS)
   }
 
   private def getNewToken(hbaseConf: HBaseConfiguration): TokenInfo = {
@@ -122,18 +120,19 @@ final class HBaseCredentialsManager private() extends Logging {
     val tokenIdentifier = token.decodeIdentifier()
     val expireTime =
       expectedExpireTime(tokenIdentifier.getIssueDate, tokenIdentifier.getExpirationDate)
+    logInfo(s"Obtain new token with expiration time $expireTime")
     TokenInfo(expireTime, hbaseConf, token)
   }
 }
 
 object HBaseCredentialsManager {
-  private var _hbaseCredentialManager: HBaseCredentialsManager = null
+  @volatile private var _hbaseCredentialManager: HBaseCredentialsManager = null
 
   /**
    * Get an instance of [[HBaseCredentialsManager]], this is a singleton to make sure only one
    * instance is created.
    */
-  def hbaseCredentialManager: HBaseCredentialsManager = {
+  def get(): HBaseCredentialsManager = {
     if (_hbaseCredentialManager == null) {
       HBaseCredentialsManager.synchronized {
         if (_hbaseCredentialManager == null) {
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
@@ -1,0 +1,148 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.execution.datasources.hbase
+
+import java.util.concurrent.{Executors, TimeUnit}
+
+import scala.collection.mutable
+import scala.language.existentials
+import scala.util.control.NonFatal
+
+import org.apache.hadoop.hbase.HBaseConfiguration
+import org.apache.hadoop.hbase.security.token.TokenUtil
+import org.apache.hadoop.security.Credentials
+import org.apache.hadoop.security.token.{Token, TokenIdentifier}
+
+import org.apache.spark.util.{ThreadUtils, Utils}
+
+final class HBaseCredentialsManager private() extends Logging {
+  private case class TokenInfo(
+      expireTime: Long,
+      conf: HBaseConfiguration,
+      token: Token[_ <: TokenIdentifier])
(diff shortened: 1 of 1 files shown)


weiqingy:
Have you tested this in the secure cluster?

jerryshao:
No, I don't have a secure cluster.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -0,0 +1,148 @@
...
+
+  private val tokenUpdater =
+    Executors.newSingleThreadScheduledExecutor(
+      ThreadUtils.namedThreadFactory("HBase Tokens Refresh Thread"))
+
+  private val tokenUpdateRunnable = new Runnable {
+    override def run(): Unit = Utils.logUncaughtExceptions(updateTokensIfRequired())
+  }
+
+  tokenUpdater.schedule(tokenUpdateRunnable, nextRefresh, TimeUnit.MILLISECONDS)
```
mridulm: In case there are exceptions in updateTokensIfRequired(), we stop updating tokens.
Why not use scheduleAtFixedRate or scheduleWithFixedDelay ? (instead of explicitly rescheduling - since the delay is constant).

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -0,0 +1,148 @@
...
+          tokenInfo.expireTime <= currTime
+        }
+
+        if (tokensShouldUpdate.isEmpty) {
+          logDebug(s"No token requires update now $currTime")
+        } else {
+          // Update all the expect to be expired tokens
+          val updatedTokens = tokensShouldUpdate.map { case (cluster, tokenInfo) =>
+            logInfo(s"Update token for cluster $cluster")
+            (cluster, getNewToken(tokenInfo.conf))
```
mridulm: As above, we should move the `getNewToken` outside the `synchronized` block; and apply `updatedTokens` to `tokensMap` within the `this.synchronized` block again.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -0,0 +1,148 @@
...
+   */
+  def hbaseCredentialManager: HBaseCredentialsManager = {
+    if (_hbaseCredentialManager == null) {
+      HBaseCredentialsManager.synchronized {
+        if (_hbaseCredentialManager == null) {
+          _hbaseCredentialManager = new HBaseCredentialsManager
+        }
+      }
+    }
+    _hbaseCredentialManager
```
mridulm: This is broken (please look up "double checked locking" for details); there are multiple ways to fix this (make method synchronized, always initialize, use volatile, etc - each with performance tradeoff).

If the cost of a thread is not too high, I would prefer always initializing `_hbaseCredentialManager -> hbaseCredentialManager` and removing the `def`

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -0,0 +1,148 @@
...
+      token: Token[_ <: TokenIdentifier])
+  private val tokensMap = new mutable.HashMap[String, TokenInfo]
+
+  // We assume token expiration time should be no less than 10 minutes.
+  private val nextRefresh = TimeUnit.MINUTES.toMillis(10)
+
+  private val tokenUpdater =
+    Executors.newSingleThreadScheduledExecutor(
+      ThreadUtils.namedThreadFactory("HBase Tokens Refresh Thread"))
+
```
mridulm: This executor should be using daemon threads.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -0,0 +1,148 @@
...
+      hbaseCluster: String,
+      hbaseConf: HBaseConfiguration): Credentials = synchronized {
+    val credentials = new Credentials()
+
+    // If token is existed and not expired, directly return the Credentials with tokens added in.
+    if (tokensMap.contains(hbaseCluster) && !isTokenExpired(tokensMap(hbaseCluster).expireTime)) {
+      credentials.addToken(tokensMap(hbaseCluster).token.getService, tokensMap(hbaseCluster).token)
+    } else {
+      // Acquire a new token if not existed or old one is expired.
+      val tokenInfo = getNewToken(hbaseConf)
```
mridulm: `getNewToken` requires an rpc call - we should do it outside the synchronized block.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -0,0 +1,148 @@
...
+   * Get HBase credential from specified cluster name.
+   */
+  def getCredentialsForCluster(
+      hbaseCluster: String,
+      hbaseConf: HBaseConfiguration): Credentials = synchronized {
+    val credentials = new Credentials()
+
+    // If token is existed and not expired, directly return the Credentials with tokens added in.
+    if (tokensMap.contains(hbaseCluster) && !isTokenExpired(tokensMap(hbaseCluster).expireTime)) {
+      credentials.addToken(tokensMap(hbaseCluster).token.getService, tokensMap(hbaseCluster).token)
```
mridulm: Performance nit:
* Instead of repeatedly calling `tokensMap.contains(hbaseCluster)` and `tokensMap(hbaseCluster)`; would be better to do a tokensMap.get(hbaseCluster) and deal with the Option returned.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -0,0 +1,148 @@
...
+      // Acquire a new token if not existed or old one is expired.
+      val tokenInfo = getNewToken(hbaseConf)
+      tokensMap.put(hbaseCluster, tokenInfo)
+      logInfo(s"Obtain new token from cluster $hbaseCluster")
+
+      credentials.addToken(tokenInfo.token.getService, tokenInfo.token)
+    }
+
+    credentials
+  }
```
mridulm: I would rewrite the method as :


If we rewrite the code along lines of :

```
def getCredentialsForCluster(hbaseCluster: String, hbaseConf: HBaseConfiguration): Credentials = {
  val tokenOpt = this.synchronized {
    tokensMap.get(hbaseCluster)
  }

  val credentials = new Credentials()
  if (tokenOpt.isDefined && ! isTokenExpired(tokenOpt.get.expireTime)) {
    // copy to credentials
  } else {
    val tokenInfo = getNewToken(hbaseConf)
    this.synchronized {
      tokensMap.put(hbaseCluster, tokenInfo)
    }
    // copy to credentials
  }
  credentials
}

```

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -0,0 +1,148 @@
...
+import org.apache.hadoop.security.Credentials
+import org.apache.hadoop.security.token.{Token, TokenIdentifier}
+
+import org.apache.spark.util.{ThreadUtils, Utils}
+
+final class HBaseCredentialsManager private() extends Logging {
+  private case class TokenInfo(
+      expireTime: Long,
+      conf: HBaseConfiguration,
+      token: Token[_ <: TokenIdentifier])
```
mridulm: Why is this a case class ?

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -0,0 +1,148 @@
...
+package org.apache.spark.sql.execution.datasources.hbase
+
+import java.util.concurrent.{Executors, TimeUnit}
+
+import scala.collection.mutable
+import scala.language.existentials
+import scala.util.control.NonFatal
+
+import org.apache.hadoop.hbase.HBaseConfiguration
+import org.apache.hadoop.hbase.security.token.TokenUtil
```
mridulm: Curious why in https://github.com/hortonworks-spark/shc/pull/107/files we use reflection for TokenUtil; but here we are directly using it.
Is there change in dependency I missed ?

+CC @weiqingy

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
jerryshao: Previously I use the minimum expire time to do reschedule, and then I found that solution has some issue, so I changed to use fix rate. I will update the code.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
jerryshao: Not sure why here it is broken. What I could think of is lacking of `volatile`.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
jerryshao: This is mainly used for Token related information, like when token is expired, and conf for token renewal.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
jerryshao: `namedThreadFactory` is a daemon thread according the implementation:

```
  def namedThreadFactory(prefix: String): ThreadFactory = {
    new ThreadFactoryBuilder().setDaemon(true).setNameFormat(prefix + "-%d").build()
  }
```

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
jerryshao: My previous intention was to avoid concurrent token renewal, not just for threading issue of `HashMap`, actually to avoid the problem you mentioned above, we could use `ConcurrentHashMap` instead. But this cannot avoid two threads to renew tokens from same cluster simultaneously.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
jerryshao: I don't think reflection is required, because we always have HBase dependencies. 

I think the reason why @weiqingy uses reflection mainly due to Spark's `HBaseCredentailProvider` implementation.

weiqingy: As @jerryshao mentioned, in the PR#107, it wanted to do what Spark's HBaseCredentailProvider does.

weiqingy:
When invoking `HBaseCredentialsManager.get().getCredentialsForCluster` in `HBaseRelation`
```
val credentials = {
    val clusterIdentifier = hbaseConf.get("zookeeper.znode.parent", "") +
      hbaseConf.get("hbase.zookeeper.quorum", "") +
      hbaseConf.get("hbase.zookeeper.property.clientPort", "")

    HBaseCredentialsManager.get()
      .getCredentialsForCluster(clusterIdentifier, hbaseConf)
  }
```

Then run the current unit tests, got the errors:
```
	at org.apache.hadoop.ipc.RemoteException.instantiateException(RemoteException.java:106)
	at org.apache.hadoop.ipc.RemoteException.unwrapRemoteException(RemoteException.java:95)
	at org.apache.hadoop.hbase.protobuf.ProtobufUtil.getRemoteException(ProtobufUtil.java:325)
	at org.apache.hadoop.hbase.protobuf.ProtobufUtil.execService(ProtobufUtil.java:1622)
	at org.apache.hadoop.hbase.ipc.RegionCoprocessorRpcChannel$1.call(RegionCoprocessorRpcChannel.java:92)
	at org.apache.hadoop.hbase.ipc.RegionCoprocessorRpcChannel$1.call(RegionCoprocessorRpcChannel.java:89)
	at org.apache.hadoop.hbase.client.RpcRetryingCaller.callWithRetries(RpcRetryingCaller.java:126)
	at org.apache.hadoop.hbase.ipc.RegionCoprocessorRpcChannel.callExecService(RegionCoprocessorRpcChannel.java:95)
	at org.apache.hadoop.hbase.ipc.CoprocessorRpcChannel.callBlockingMethod(CoprocessorRpcChannel.java:73)
	at org.apache.hadoop.hbase.protobuf.generated.AuthenticationProtos$AuthenticationService$BlockingStub.getAuthenticationToken(AuthenticationProtos.java:4512)
	at org.apache.hadoop.hbase.security.token.TokenUtil.obtainToken(TokenUtil.java:86)
	at org.apache.hadoop.hbase.security.token.TokenUtil.obtainToken(TokenUtil.java:69)
	at org.apache.spark.sql.execution.datasources.hbase.HBaseCredentialsManager.org$apache$spark$sql$execution$datasources$hbase$HBaseCredentialsManager$$getNewToken(HBaseCredentialsManager.scala:119)
	at org.apache.spark.sql.execution.datasources.hbase.HBaseCredentialsManager.getCredentialsForCluster(HBaseCredentialsManager.scala:68)
	at org.apache.spark.sql.execution.datasources.hbase.HBaseRelation.<init>(HBaseRelation.scala:127)
	at org.apache.spark.sql.execution.datasources.hbase.DefaultSource.createRelation(HBaseRelation.scala:62)
	at org.apache.spark.sql.execution.datasources.DataSource.write(DataSource.scala:426)
	at org.apache.spark.sql.DataFrameWriter.save(DataFrameWriter.scala:215)
	at org.apache.spark.sql.DefaultSourceSuite$$anonfun$1.apply$mcV$sp(DefaultSourceSuite.scala:89)
	at org.apache.spark.sql.DefaultSourceSuite$$anonfun$1.apply(DefaultSourceSuite.scala:78)
	at org.apache.spark.sql.DefaultSourceSuite$$anonfun$1.apply(DefaultSourceSuite.scala:78)
```

Need to figure out if the error only happens in unit or not tests since the tests are running in the minicluster instead of real clusters. The new secure cluster was just created in Jenkin this morning, but not ready for using. 

(BTW, I changed the function signature from `getCredentialsForCluster(String, HBaseConfiguration)` to `getCredentialsForCluster(String, Configuration)`, although it should have nothing to do with the errors.)

mridulm:
@weiqingy Is there some portion of the stack trace missing ?
The 'reason' for the exception is not seen - which will tell us why the token acquisition failed.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
mridulm: Ah, I did not notice that - thanks for clarifying !

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
mridulm: We cannot avoid concurrent token renewal - there is a daemon thread doing that in parallel; which can potentially block any use of getCredentialsForCluster indefinitely (and vice versa).

Using ConcurrentHashMap can be fine too - though it will make the code more verbose without much benefit.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
mridulm: Why it is broken has to do with guarantees exposed by java's memory model - use of volatile is a way to avoid DCL (which works since java 1.5).

Given that this is scala code; the best solution is to actually use a `lazy val` (which does the right thing) - should have mentioned that initially - and get rid of all the code in the object.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
mridulm: Interesting, then we should ideally fix PR 107 (sometime in future) to avoid reflection if we always have this dependency in shc.

weiqingy:
I updated the stack trace.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
weiqingy: PR 107 was not merged in SHC, just closed without being merged in.

mridulm:
@weiqingy Just curious - is there a client/server version mismatch ?
A cursory google search seemed to indicate that ...

weiqingy:
@mridulm I am debugging this. It seems we need to specify `hbase.coprocessor.region.classes`
`to org.apache.hadoop.hbase.security.token.TokenProvider`, so I was trying to setup the minicluster as following:
```
private[spark] var htu = new HBaseTestingUtility
....
UserProvider.setUserProviderForTesting(htu.getConfiguration, classOf[HadoopSecurityEnabledUserProviderForTesting])
SecureTestUtil.enableSecurity(htu.getConfiguration)
htu.getConfiguration.set("hbase.coprocessor.region.classes", "org.apache.hadoop.hbase.security.token.TokenProvider")
```
but the test is still hanged `TokenUtil.obtainToken(hbaseConf)` hangs at [getAuthenticationToken](https://github.com/apache/hbase/blob/master/hbase-server/src/main/java/org/apache/hadoop/hbase/security/token/TokenUtil.java#L71).

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
jerryshao: No special reason, it can be a regular class. Let me change it.

jerryshao:
@weiqingy I tested in real cluster and it worked. I'm wondering if you're going to test on mini hbase cluster, then how do you enable kerberos things?

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
mridulm: Sounds good, thx.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
mridulm: Thanks !
The reason for asking is, I had to remove a bunch of `case class` in SHC in the past iirc.
Initially it starts as a case class for ease of coding, then it becomes mutable, and pretty soon it starts breaking `case class` contract ... if we do need it to be a case class, then it is fine - else would be good to avoid its use.

weiqingy:
@jerryshao I just talked to Hbase team. There are some configurations need to set up in the minicluster. So right now, yes,  let's just do the testing in the real cluster.

But If you invoke in HBaseRelation, unit tests will fail.
```
val credentials = {
    val clusterIdentifier = hbaseConf.get("zookeeper.znode.parent", "") +
      hbaseConf.get("hbase.zookeeper.quorum", "") +
      hbaseConf.get("hbase.zookeeper.property.clientPort", "")

    HBaseCredentialsManager.get()
      .getCredentialsForCluster(clusterIdentifier, hbaseConf)
  }
```

weiqingy:
@jerryshao Could you share your secure cluster? Just checked with Yesha, the secure cluster build failed again.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -0,0 +1,140 @@
...
+    override def run(): Unit = Utils.logUncaughtExceptions(updateTokensIfRequired())
+  }
+
+  tokenUpdater.scheduleAtFixedRate(
+    tokenUpdateRunnable, nextRefresh, nextRefresh, TimeUnit.MILLISECONDS)
+
+  /**
+   * Get HBase credential from specified cluster name.
+   */
+  def getCredentialsForCluster(
```
weiqingy: Could we change getCredentialsForCluster method to have signature : `def getCredentialsForCluster(conf: Configuration): Credentials` as mentioned in [here].(https://github.com/weiqingy/shc/commit/e8c60120cd42d1500dae1ba6f641ce0f0bf7c42e#diff-7c2c495c4b6d1b09f05856320c914948R125)

weiqingy:
@jerryshao Could we also add a API like `credentialsRequired`? Since current minicluster for unit tests does not support secure stuff, the API `credentialsRequired` can be used to skip the logic about secure thing.

jerryshao:
Ping @mridulm , can you please review again?

@weiqingy , you could always add missing features to this code after get merged. Because I'm not familiar with other codes, so the design may not be so suitable for other parts to leverage it.

jerryshao:
@weiqingy I'm going to merge this to not block on the following works.

jerryshao:
@weiqingy is branch-2.1 the right branch to land in patch?

weiqingy:
Yes. branch-2.1.

PR #111, opened 2017-03-30
Ensure a hbase cluster issue does not cause other token updates to fail

mridulm:
+CC @jerryshao

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
@@ -98,30 +98,34 @@ final class HBaseCredentialsManager private() extends Logging {
   }
 
   private def updateTokensIfRequired(): Unit = {
-    try {
-      val currTime = System.currentTimeMillis()
+    val currTime = System.currentTimeMillis()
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -98,30 +98,34 @@ final class HBaseCredentialsManager private() extends Logging {
...
 
-        this.synchronized {
-          updatedTokens.foreach { kv => tokensMap.put(kv._1, kv._2) }
-        }
+      this.synchronized {
+        updatedTokens.foreach { kv => tokensMap.put(kv._1, kv._2) }
       }
-    } catch {
-      case NonFatal(e) =>
-        logWarning("Error while trying to fetch tokens from HBase cluster", e)
```
mridulm: Essentially, I moved the catch NonFatal to block around getNewToken()

jerryshao: Thanks @mridulm , just saw your comments. 

Yes, indeed we should do like this.

mridulm: Thx for merging it !

PR #116, opened 2017-03-31
Delegate to specific coder when scanning for Strings, Array[Byte], etc

rayokota:

The calculation of the range for the scanner is not correct for types String, Array[Byte], Byte, UTF8String. The serialized representations should be obtained from the coder. In the case of PrimitiveType and Phoenix, the serializations happen to match the code in ScanRange.scala. However, I have a custom coder that does not have the same serialization.

## How was this patch tested?

All existing unit tests still pass


weiqingy:
Thanks @rayokota . Merged to master/branch-1.6/branch-2.0/branch-2.1.

PR #118, opened 2017-04-05
Always fetch authentication tokens in Kerberos granted UGI

jerryshao:
CC @mridulm @bikassaha

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
@@ -26,6 +26,7 @@ import scala.util.control.NonFatal
 
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.hbase.security.token.TokenUtil
+import org.apache.hadoop.io.Text
 import org.apache.hadoop.security.{Credentials, UserGroupInformation}
 import org.apache.hadoop.security.token.{Token, TokenIdentifier}
 
@@ -80,7 +81,7 @@ final class HBaseCredentialsManager private(sparkConf: SparkConf) extends Loggin
       }
       logInfo(s"Obtain new token for cluster $identifier")
 
-      credentials.addToken(tokenInfo.token.getService, tokenInfo.token)
+      credentials.addToken(new Text(identifier), tokenInfo.token)
     }
 
     credentials
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
@@ -17,6 +17,7 @@
 
 package org.apache.spark.sql.execution.datasources.hbase
 
+import java.security.PrivilegedExceptionAction
 import java.util.concurrent.{Executors, TimeUnit}
 
 import scala.collection.mutable
@@ -28,9 +29,10 @@ import org.apache.hadoop.hbase.security.token.TokenUtil
 import org.apache.hadoop.security.{Credentials, UserGroupInformation}
 import org.apache.hadoop.security.token.{Token, TokenIdentifier}
 
+import org.apache.spark.SparkConf
 import org.apache.spark.util.{ThreadUtils, Utils}
 
-final class HBaseCredentialsManager private() extends Logging {
+final class HBaseCredentialsManager private(sparkConf: SparkConf) extends Logging {
   private class TokenInfo(
       val expireTime: Long,
       val conf: Configuration,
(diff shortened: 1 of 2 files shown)


mridulm:
I am not sure if we can assume we will always have principal/keytab.
@bikassaha any comments ?

jerryshao:
If not, then I don't know there's a way to address it.

mridulm:
@jerryshao I am trying to find a workaround ... if that does not work, we will go with this as fallback

jerryshao:
@mridulm I think the keypoint is to get a UGI with only tgt granted, so HBase client will honor this tgt to get new tokens.

jerryshao:
Let me merge this PR first, @weiqingy you need to change the code of #120  based on this.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -44,6 +47,11 @@ final class HBaseCredentialsManager private() extends Logging {
     Executors.newSingleThreadScheduledExecutor(
       ThreadUtils.namedThreadFactory("HBase Tokens Refresh Thread"))
 
+  require(sparkConf.contains("spark.yarn.principal"))
+  require(sparkConf.contains("spark.yarn.keytab"))
+  private val principal = sparkConf.get("spark.yarn.principal")
+  private val keytab = sparkConf.get("spark.yarn.keytab")
+
```
mridulm: Based on comments by @bikassaha , we are supposed to have shc specific principal/keytab and falling back to spark config if unavailable.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -73,7 +81,7 @@ final class HBaseCredentialsManager private() extends Logging {
       }
       logInfo(s"Obtain new token for cluster $identifier")
 
-      credentials.addToken(tokenInfo.token.getService, tokenInfo.token)
+      credentials.addToken(new Text(identifier), tokenInfo.token)
```
mridulm: This is incorrect, as discussed yesterday, this must be `token.getService`

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -130,12 +138,17 @@ final class HBaseCredentialsManager private() extends Logging {
...
-    new TokenInfo(expireTime, conf, token)
+    val kerberosLoggedUGI = UserGroupInformation.loginUserFromKeytabAndReturnUGI(principal, keytab)
+    kerberosLoggedUGI.doAs(new PrivilegedExceptionAction[TokenInfo] {
+      override def run(): TokenInfo = {
+        val token = TokenUtil.obtainToken(conf)
+        val tokenIdentifier = token.decodeIdentifier()
+        val expireTime =
+          expectedExpireTime(tokenIdentifier.getIssueDate, tokenIdentifier.getExpirationDate)
+        logInfo(s"Obtain new token with expiration time $expireTime")
+        new TokenInfo(expireTime, conf, token)
```
mridulm: @weiqingy Please make sure we change this to accommodate review comments from @bikassaha about not overloading expireTime; but rely on renewTime and expireTokenInfoTime.
Also about caching the serialized token and not token itself.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
```diff
@@ -150,5 +163,16 @@ final class HBaseCredentialsManager private() extends Logging {
...
+
+  def get(sparkConf: SparkConf): HBaseCredentialsManager = {
+    if (_hbaseCredentialManager == null) {
+      HBaseCredentialsManager.synchronized {
+        if (_hbaseCredentialManager == null) {
+          _hbaseCredentialManager = new HBaseCredentialsManager(sparkConf)
+        }
+      }
+    }
+    _hbaseCredentialManager
```
mridulm: This change is not required - lazy val does the exact same thing without all the code.
In addition, HBaseCredentialsManager must be created/initialized only when configuration enables it.

weiqingy:
@jerryshao I just reverted the PR, but we can always merge it in after #120 is in and the comments are resolved.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala
weiqingy: Sure. I will update the PR to caching the serialized token instead.

jerryshao:
Sure, thanks.

weiqingy:
Hi, @jerryshao PR#120 is merged. Could you please commit the PR again? Thanks.

jerryshao:
Will do it. Currently stuck on the some Spark issues, will do it after that.

PR #120, opened 2017-04-06
Modify the implementation of Supporting multiple Hbase secure clusters

weiqingy:

The PR is based on all of the previous [code changes and review comments](https://github.com/weiqingy/shc/commit/48c00ee75ad1570663e35dc31f4534ca1b3f879d) including [Mridul's PR](https://github.com/weiqingy/shc/pull/1). 

Changes: 
Clean code related to logs; Move serialize/deserialize to SHCCredientialManager.

## How was this patch tested?

Passed build. Passed existing unit tests .


weiqingy:
Could you please help to review this PR? @mridulm @bikassaha @jerryshao Thanks!

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+import org.apache.spark.util.{ThreadUtils, Utils}
+
+final class SHCCredentialsManager private() extends Logging {
+  private class TokenInfo(
+    val expireTime: Long,
+    val issueTime: Long,
+    val conf: Configuration,
+    val token: Token[_ <: TokenIdentifier]) {
+
+    val isTokenInfoExpired: Boolean = {
```
jerryshao: Does it really worked? Shouldn't be `def`? Also applies to the below code.

weiqingy: Yes. We have tested this [one](https://github.com/mridulm/shc/blob/1d25163c97d742f489a452925e15c00ab939fd3b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseCredentialsManager.scala#L49) in clusters many times? Is there any issue here?

jerryshao: This is a immutable value initialized in constructing, using this to check if token is expired will always be 'false'...

mridulm: @jerryshao is right, this should be a def.
The `System.currentTimeMillis()` should be the current time at invocation time - not at construction time.

I think this has changed from `val tokenInfoExpireTime: Long` (which did not have `System.currentTimeMillis() >=`) to including the time check - which requires it to be a `def`

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+    val identifier = clusterIdentifier(conf)
+
+    val tokenOpt = this.synchronized {
+      tokensMap.get(identifier)
+    }
+
+    // If token is existed and not expired, directly return the Credentials with tokens added in.
+    if (tokenOpt.isDefined && !tokenOpt.get.isTokenInfoExpired) {
+      credentials.addToken(tokenOpt.get.token.getService, tokenOpt.get.token)
+      logInfo(s"Use existing token for on-demand cluster $identifier")
```
mridulm: nit: Should we make this `logDebug` instead ?

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    UserGroupInformation.getCurrentUser.addCredentials(credentials)
+    SHCCredentialsManager.addLogs("Driver", credentials) // Only driver invokes getCredentialsForCluster in SHC
+    SHCCredentialsManager.serialize(credentials)
+  }
+
+  def isCredentialsRequired(conf: Configuration): Boolean = {
+    UserGroupInformation.isSecurityEnabled &&
+      conf.get("hbase.security.authentication") == "kerberos"
```
mridulm: I thought we wanted this to be controlled by some flag ? To disable SHCCredentialsManager entirely ?
(in addition to the conditions above ofcourse)

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+      }
+
+      logInfo(s"getCredentialsForCluster: Obtained new token with expiration time" +
+        s" ${new Date(tokenInfo.expireTime)} and refresh time ${new Date(tokenInfo.refreshTime)} " +
+        s"for cluster $identifier")
+
+      credentials.addToken(tokenInfo.token.getService, tokenInfo.token)
+    }
+
+    UserGroupInformation.getCurrentUser.addCredentials(credentials)
```
mridulm: I thought we are not doing this ? Based on the latest test ?

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+    new TokenInfo(expireTime, issueTime, conf, token)
+  }
+
+  private def clusterIdentifier(conf: Configuration): String = {
+    require(conf.get("zookeeper.znode.parent") != null &&
+      conf.get("hbase.zookeeper.quorum") != null &&
+      conf.get("hbase.zookeeper.property.clientPort") != null)
+
+    conf.get("zookeeper.znode.parent") + "-"
+      conf.get("hbase.zookeeper.quorum") + "-"
```
mridulm: nit: Ideally, we should use a delimiter which is not valid hostname. perhaps some punctuation character like '!' ?

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: Terrible error！Fixed

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
mridulm: Note that when @jerryshao makes the principal/keytab based change, this will be required : but token acquisition will happen with UGI call to auth user.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
jerryshao: I could not figure out a proper delimiter, using "!" is little weird.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+    val credentials = new Credentials()
+    val identifier = clusterIdentifier(conf)
+
+    val tokenOpt = this.synchronized {
+      tokensMap.get(identifier)
+    }
+
+    // If token is existed and not expired, directly return the Credentials with tokens added in.
+    if (tokenOpt.isDefined && !tokenOpt.get.isTokenInfoExpired) {
+      credentials.addToken(tokenOpt.get.token.getService, tokenOpt.get.token)
```
jerryshao: `tokenOpt.get.token.getService` be better to change to `identifier`. I would doubt `tokenOpt.get.token.getService` will be duplicated if multiple hbase tokens are added. For the safe we could use `identifier` instead.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
mridulm: The idea is to ensure the delimiter we use cannot occur in the values we are delimiting normally - in order to avoid situations where : <str1> <delim> <str2> == <str3><delim><str4>
If delimiter can occur within str, then it can potentially lead to false collusion (this might look like a theoretical point until it actually hits you, having faced it in past, I am very paranoid about delimiters !)


Note that this key is not exposed externally - so it does not need to be pretty.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
mridulm: Service would be the cluster id right ? Shouldn't that not be unique ?
I thought hbase looked up tokens based on service name for auth ?

mridulm: This is based on a cursory reading of org.apache.hadoop.hbase.security.token.TokenUtil btw !

jerryshao: I"m not sure, if you're sure about the uniqueness of `getService` then I'm totally fine with it. The Hive patch here (https://issues.apache.org/jira/secure/attachment/12681574/HIVE-8874.1.patch) made me think of this 😄 .

mridulm: I am not sure if I missed something here - but isn't the hive patch not doing the same ? (`getService` for key in credential ?)
I see that in `addHBaseDelegationToken()` in that patch.
`getAuthToken()` token is doing what `TokenUtil` is also doing - key is the clusterId (and so not a string we customize).

My assumption is that, from `Credentials`, based on the clusterId the approriate token is looked up : hence the key cannot change : but I am yet to dig into hbase to validate this !

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: Yes. :)

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: This has been removed in the updated PR.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: Yes, a configure parameter has been added to control this.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: "#" has been used. :)

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
mridulm: A cursory look at org.apache.hadoop.hbase.ipc.RpcConnection seems to indicate this is the case.
Take a look at https://github.com/apache/hbase/blob/4e77b18da2515a14772a456f408ee34376a3c71f/hbase-client/src/main/java/org/apache/hadoop/hbase/ipc/RpcConnection.java#L106

jerryshao: @mridulm , I was thinking of this line `return new AuthenticationTokenSelector().selectToken(new Text(clusterId), user.getUGI().getTokens());` in hive patch. I think you're right， maybe I'm worrying too much.

mridulm: My understanding from hbase code is that, for token auth (DIGEST mechanism), it looks up based on the cluster id.
For kerberos auth (when not using tokens, and auth enabled), then it is based on ugi.

weiqingy:
Tested the JoinTablesFrom2Clusters (without the loop of accessing Hbase tables) example in a single cluster, worked well.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+        updatedTokens.foreach { kv => tokensMap.put(kv._1, kv._2) }
+      }
+    }
+  }
+
+  private def getNewToken(conf: Configuration): TokenInfo = {
+    val token = TokenUtil.obtainToken(conf)
+    val tokenIdentifier = token.decodeIdentifier()
+    val expireTime = tokenIdentifier.getExpirationDate
+    val issueTime = tokenIdentifier.getIssueDate
```
bikassaha: Per email from Jerry this should also add the token to the UGI.

mridulm: We have to wait for Jerry's patch to make it in before we can do that.
Until his patch is committed (support for principal/keytab), **not** adding to UGI is the only way we can acquire tokens for executors :-)

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+    }
+  }
+
+  // for debug
+  def addLogs(component: String, credentials: Credentials): Unit = {
+    logInfo(s"$component: Obtain credentials with minimum expiration date of " +
+      s"tokens ${getMinimumExpirationDates(credentials).getOrElse(-1)}")
+  }
+
+  private def getMinimumExpirationDates (credentials: Credentials): Option[Long] = {
```
jerryshao: remove white space after "Dates".

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseRelation.scala
```diff
@@ -328,28 +327,6 @@ object HBaseRelation {
...
-      val dataStream = new DataInputStream(byteStream)
-      val credentials = new Credentials()
-      credentials.readTokenStorageStream(dataStream)
-      credentials
-    } else {
-      null
-    }
-  }
+  // Set false to disable SHCCredentialsManager
+  val EnableSHCCredentialsManager = "true"
```
mridulm: This should be obtained from config and not hard coded.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+import org.apache.hadoop.security.token.{Token, TokenIdentifier}
+
+import org.apache.spark.util.{ThreadUtils, Utils}
+
+final class SHCCredentialsManager private() extends Logging {
+  private class TokenInfo(
+    val expireTime: Long,
+    val issueTime: Long,
+    val conf: Configuration,
+    val token: Token[_ <: TokenIdentifier]) {
```
mridulm: Based on feedback yesterday, instead of keeping only token and repeatedly serializing it, we need to also add serializedToken here - which is created/added as part of `getNewToken`
token will be used to add to driver credentials and getService.
serializedToken will be used to return as Array[Byte] from `getCredentialsForCluster` to avoid repeatedly having to invoke `SHCCredentialsManager.serialize(credentials)` for each invocation of `getCredentialsForCluster`

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+    conf.getBoolean(SparkHBaseConf.credentialsManagerEnabled, true) &&
+      UserGroupInformation.isSecurityEnabled &&
+      conf.get("hbase.security.authentication") == "kerberos"
+  }
+
+  private def updateTokensIfRequired(): Unit = {
+    val currTime = System.currentTimeMillis()
+
+    // Filter out all the tokens should be re-issued.
+    val tokensShouldUpdate = this.synchronized {
```
mridulm: nit: `tokensShouldUpdate` -> `tokensToUpdate`

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+      conf.get("hbase.zookeeper.property.clientPort")
+  }
+}
+
+object SHCCredentialsManager extends Logging {
+  lazy val manager = new  SHCCredentialsManager
+
+  private val expireTimeFraction = 0.95
+  private val refreshTimeFraction = 0.6
+  private val refreshDurationMins = 10
```
mridulm: nit: Ideally, these should have been configurable from sparkConf - in debugging efforts I left them as constants in code I was directly modifying to test.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: Done.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+  def addLogs(component: String, credentials: Credentials): Unit = {
+    logInfo(s"$component: Obtain credentials with minimum expiration date of " +
+      s"tokens ${getMinimumExpirationDates(credentials).getOrElse(-1)}")
+  }
+
+  private def getMinimumExpirationDates(credentials: Credentials): Option[Long] = {
+    val expirationDates = credentials.getAllTokens.asScala
+      .filter(_.decodeIdentifier().isInstanceOf[AuthenticationTokenIdentifier])
+      .map { t =>
+        val identifier = t.decodeIdentifier().asInstanceOf[AuthenticationTokenIdentifier]
```
mridulm: `decodeIdentifier` looks slightly expensive - we are doing two decode's here - one in `filter` and another in `map`; can we minimize it ?

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseRelation.scala
weiqingy: What I thought was: users can set the value of EnableSHCCredentialsManager in "`options`" 
```
sc.parallelize(oldData).toDF.write.options(
      Map(HBaseRelation.EnableSHCCredentialsManager -> "false", HBaseTableCatalog.tableCatalog -> catalog, HBaseTableCatalog.newTable -> "5"))
      .format("org.apache.spark.sql.execution.datasources.hbase")
      .save()
```
For example, users can set `HBaseRelation.TIMESTAMP` in [here](https://github.com/hortonworks-spark/shc/blob/master/core/src/test/scala/org/apache/spark/sql/DefaultSourceSuite.scala#L287 )

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseRelation.scala
```diff
@@ -105,6 +105,7 @@ case class HBaseRelation(
         val conf = HBaseConfiguration.create
         hBaseConfiguration.foreach(_.foreach(e => conf.set(e._1, e._2)))
         hBaseConfigFile.foreach(e => conf.set(e._1, e._2))
+        conf.setBoolean(SparkHBaseConf.credentialsManagerEnabled, enableCredsManager.getOrElse(true))
```
weiqingy: If users do not specify HBaseRelation.EnableSHCCredentialsManager, then its default value is true.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseRelation.scala
weiqingy: The current code is confusing. Let me change it. Maybe we can just use `SparkHBaseConf.credentialsManagerEnabled`

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+        s"for cluster $identifier")
+
+      credentials.addToken(tokenInfo.token.getService, tokenInfo.token)
+    }
+
+    SHCCredentialsManager.addLogs("Driver", credentials) // Only driver invokes getCredentialsForCluster
+    SHCCredentialsManager.serialize(credentials)
+  }
+
+  def isCredentialsRequired(conf: Configuration): Boolean = {
```
bikassaha: Should be done only once.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseRelation.scala
```diff
@@ -117,15 +118,12 @@ case class HBaseRelation(
 
   def hbaseConf = wrappedConf.value
 
-  val serializedCredentials = {
-    if (HBaseCredentialsManager.manager.isCredentialsRequired(hbaseConf)) {
-      val credentials = HBaseCredentialsManager.manager.getCredentialsForCluster(hbaseConf)
-      UserGroupInformation.getCurrentUser.addCredentials(credentials)
-      HBaseRelation.serialize(credentials)
+  val serializedCredentials =
+    if (SHCCredentialsManager.manager.isCredentialsRequired(hbaseConf)) {
```
bikassaha: Should not be compute from conf each time.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseRelation.scala
```diff
@@ -236,8 +234,9 @@ case class HBaseRelation(
 
     rdd.mapPartitions(iter => {
       if (null != serializedCredentials) {
-        UserGroupInformation.getCurrentUser
-          .addCredentials(HBaseRelation.deserialize(serializedCredentials))
+        val creds = SHCCredentialsManager.deserialize(serializedCredentials)
+        SHCCredentialsManager.addLogs("Task", creds) // for debug
```
bikassaha: Dont log inside credentials manager because the log will be confusing to read.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseRelation.scala
```diff
@@ -236,8 +234,9 @@ case class HBaseRelation(
...
+        UserGroupInformation.getCurrentUser.addCredentials(creds)
       }
       iter.map(convertToPut)
     }).saveAsNewAPIHadoopDataset(job.getConfiguration)
```
bikassaha: Does saveAsNew have a repartition inside it.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+final class SHCCredentialsManager private() extends Logging {
+  private class TokenInfo(
+    val expireTime: Long,
+    val issueTime: Long,
+    val conf: Configuration,
+    val token: Token[_ <: TokenIdentifier]) {
+
+    def isTokenInfoExpired: Boolean = {
+      System.currentTimeMillis() >=
+        ((expireTime - issueTime) * SHCCredentialsManager.expireTimeFraction + issueTime).toLong
```
bikassaha: No need for a fraction here too. Expire should return true if its really expired.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
bikassaha: This should be serialized credentials

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+      tokensMap.get(identifier)
+    }
+
+    // If token is existed and not expired, directly return the Credentials with tokens added in.
+    if (tokenOpt.isDefined && !tokenOpt.get.isTokenInfoExpired) {
+      credentials.addToken(tokenOpt.get.token.getService, tokenOpt.get.token)
+      logDebug(s"Use existing token for on-demand cluster $identifier")
+    } else {
+
+      logInfo(s"getCredentialsForCluster: Obtaining new token for cluster $identifier")
```
bikassaha: Break into 2 parts. First time token acquisition for new cluster. log info. 
Getting new token for existing cluster. Should never happen if refresh is working. log warn.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+      conf.get("hbase.security.authentication") == "kerberos"
+  }
+
+  private def updateTokensIfRequired(): Unit = {
+    val currTime = System.currentTimeMillis()
+
+    // Filter out all the tokens should be re-issued.
+    val tokensShouldUpdate = this.synchronized {
+      tokensMap.filter { case (_, tokenInfo) => tokenInfo.refreshTime <= currTime }
+    }
```
bikassaha: This can have bug based on values of refresh thread interval, expire time and refresh time of tokens. Perhaps we should run refresh thread every 30 mins. That might resolve the previous comment about handling failures.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+  private val tokenUpdater =
+    Executors.newSingleThreadScheduledExecutor(
+      ThreadUtils.namedThreadFactory("HBase Tokens Refresh Thread"))
+
+  private val tokenUpdateRunnable = new Runnable {
+    override def run(): Unit = Utils.logUncaughtExceptions(updateTokensIfRequired())
+  }
+
+  tokenUpdater.scheduleAtFixedRate(
+    tokenUpdateRunnable, nextRefresh, nextRefresh, TimeUnit.MILLISECONDS)
```
bikassaha: How to handle failures

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+      logInfo("Refresh Thread: No tokens require update")
+    } else {
+      // Update all the expect to be expired tokens
+      val updatedTokens = tokensShouldUpdate.map { case (cluster, tokenInfo) =>
+        logInfo(s"Refresh Thread: Update token for cluster $cluster")
+
+        val token = {
+          try {
+            val tok = getNewToken(tokenInfo.conf)
+            logInfo(s"Refresh Thread: Successfully obtained token for cluster $cluster")
```
bikassaha: Could try to combine logging to reduce verbosity.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+      }
+    }
+  }
+
+  private def getNewToken(conf: Configuration): TokenInfo = {
+    val token = TokenUtil.obtainToken(conf)
+    val tokenIdentifier = token.decodeIdentifier()
+    val expireTime = tokenIdentifier.getExpirationDate
+    val issueTime = tokenIdentifier.getIssueDate
+    new TokenInfo(expireTime, issueTime, conf, token)
```
bikassaha: Cluster identifier string could be inside token info to fully define the payload.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,217 @@
...
+    tokenUpdateRunnable, nextRefresh, nextRefresh, TimeUnit.MILLISECONDS)
+
+  /**
+   * Get HBase credential from specified cluster name.
+   */
+  def getCredentialsForCluster(conf: Configuration): Array[Byte] = {
+    val credentials = new Credentials()
+    val identifier = clusterIdentifier(conf)
+
+    val tokenOpt = this.synchronized {
```
bikassaha: Instead of synchronized can we use concurrenthashmap.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseRelation.scala
weiqingy: removed.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
mridulm: The runnable should keep getting scheduled for that rate even if there are exceptions.
The actual token acquisition per cluster is within a try/catch - so that one bad cluster does not make token acquisition for other clusters fail.

There are still corner cases we dont catch - like if the first cluster's token acquisition blocks "for a while" due to network issues, etc.
That is the reason why we have a last ditch fallback in `getCredentialsForCluster`

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseRelation.scala
weiqingy: Updated the PR for this.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: Done.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: Updated PR to get them from sparkConf.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: Yes. I have updated PR for this.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: In the compute() (task side), we need` UserGroupInformation.getCurrentUser.addCredentials(credentials)`, so we need `serializedCredentials` instead of `serializedToken` return as Array[Byte] from getCredentialsForCluster.

weiqingy: Done: a8d7e09

weiqingy: How about serializedToken instead of serializedCredentials? There is only one token inside the credential, so change the function from `getCredentialsForCluster` to `getTokenForCluster`.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: I think it's better to add credentials to the current UGI after Jerry's PR is merged, otherwise the testing will be failed for the current implementation.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseRelation.scala
weiqingy: Fixed.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: yeah, some unnecessary log removed.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: Done.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseRelation.scala
```diff
@@ -235,9 +234,15 @@ case class HBaseRelation(
     }
 
     rdd.mapPartitions(iter => {
-      if (null != serializedCredentials) {
-        UserGroupInformation.getCurrentUser
-          .addCredentials(HBaseRelation.deserialize(serializedCredentials))
+      if (null != serializedToken) {
+        val tok = SHCCredentialsManager.deserializeToken(serializedToken)
+        val credentials = new Credentials()
+        credentials.addToken(tok.getService, tok)
```
mridulm: IIRC @bikassaha's review comment was the this should be deserializeCredential - we can replace this block of code with `ugi.getCurrentUser.addCredentials(SHCCredentialsManager.deserializeCredential(serializedCredential)`.

Have you kept it as token for the purpose of logging the expiration timestamp ?
If yes, that might be useful - any comments @bikassaha on keeping it this way ? Now, at driver, we serialize token once : instead of repeatedly serializing credentials earlier.
The 'cost' at driver and executor should be same as with serializing Credentials as you proposed.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
```diff
@@ -54,7 +55,7 @@ private[hbase] class HBaseTableScanRDD(
   val outputs = StructType(requiredColumns.map(relation.schema(_))).toAttributes
   val columnFields = relation.splitRowKeyColumns(requiredColumns)._2
   private def sparkConf = SparkEnv.get.conf
-  val credentials = relation.serializedCredentials
+  val token = relation.serializedToken
```
mridulm: Rename this to `serializedToken` to be clear that it is in serialized form, like we do in `HBaseRelation`

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
```diff
@@ -257,8 +258,15 @@ private[hbase] class HBaseTableScanRDD(
...
-      UserGroupInformation.getCurrentUser.addCredentials(HBaseRelation.deserialize(credentials))
+    if (null != token) {
+      val tok = SHCCredentialsManager.deserializeToken(token)
+      val credentials = new Credentials()
+      credentials.addToken(tok.getService, tok)
+
+      logInfo(s"Task: Obtain token with expiration date" +
+        s" ${tok.decodeIdentifier().asInstanceOf[AuthenticationTokenIdentifier].getExpirationDate}")
+
+      UserGroupInformation.getCurrentUser.addCredentials(credentials)
```
mridulm: This block of code seems to be repeated in `HBaseRelation` and `HBaseTableScan`.

Instead of exposing `serializeToken` and `deserializeToken` methods in `SHCCredentialsManager`, we could add a single public method to process `serializedToken` - perhaps `processShcToken` ? 
And keep the implementation of what is done to generate/process the `Array[Byte]` private to `SHCCredentialsManager.
So that tomorrow, if required, we can change the impl in `SHCCredentialsManager` without modifying rest of codebase - and not exposing how we generate/process the token or credential or etc.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,209 @@
...
+final class SHCCredentialsManager private(sparkConf: SparkConf) extends Logging {
+  private class TokenInfo(
+    val expireTime: Long,
+    val issueTime: Long,
+    val refreshTime: Long,
+    val conf: Configuration,
+    val token: Token[_ <: TokenIdentifier],
+    val serializedToken: Array[Byte])
+
+  private val expireTimeFraction =sparkConf.getDouble(SparkHBaseConf.expireTimeFraction, 0.95)
```
mridulm: nit: space after `=`

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,209 @@
...
+  private val refreshTimeFraction = sparkConf.getDouble(SparkHBaseConf.refreshTimeFraction, 0.6)
+  private val refreshDurationMins = sparkConf.getInt(SparkHBaseConf.refreshDurationMins, 10)
```
mridulm: We should pull these variables into `SHCCredentialsManager` and not keep them in `TokenInfo`

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,209 @@
...
+  def getTokenForCluster(conf: Configuration): Array[Byte] = {
+    // var token: Token[_ <: TokenIdentifier] = null
+    var serializedToken: Array[Byte] = null
+    val identifier = clusterIdentifier(conf)
+
+    val tokenInfoOpt = this.synchronized {
+      tokensMap.get(identifier)
+    }
+
+    var needNewToken = true
```
mridulm: Instead of `var`, we can make it the result of the body.

Like:
```
val needNewToken = {
  if (condition1){
    if (c1) {
      ...
      true
    } else {
     ...
     false
    }
  } else {
    ...
    true
  }
}
```

This prevents future accidental modifications to the variable.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,209 @@
...
+    // TODO: the code below and the commented variable 'token' above will be rewritten or uncommented in the following PR.
+    /* val credentials = new Credentials()
+    credentials.addToken(token.getService, token)
+    UserGroupInformation.getCurrentUser.addCredentials(credentials)
+    */
+
+    serializedToken
+  }
+
+  def isCredentialsRequired(conf: Configuration): Boolean = {
```
mridulm: Instead of exposing this method, why not simply enforce it in `getTokenForCluster` to return null if it is false ?
Keeps things simpler for the developer - always call `manager.getTokenForCluster` : null if credentials/token absent, else actual value.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,209 @@
...
+    sparkConf.getBoolean(SparkHBaseConf.credentialsManagerEnabled, true) &&
+      UserGroupInformation.isSecurityEnabled &&
```
mridulm: `sparkConf.getBoolean && UserGroupInformation.isSecurityEnabled` should be computed once as a class instance variable and reused instead of invoking each time.
With perhaps a log message after its computation at the class initialization time - so that users know whether we have enabled credential updates or not.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: Actually it's been in `SHCCredentialsManager` not in `TokenInfo`.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: yeah.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
mridulm: You are right, please ignore my comment - somehow I misread it in github UI !

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseRelation.scala
weiqingy: I chose to serialize token instead of credential for two reasons: one is to make logging easier, the other one is for performance. I suppose serializing, transporting and deserializing a token should be less expensive than a credential. Of course I can change to credential if you guys prefer, just comment and let me know.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
weiqingy: yes.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: @mridulm Yes. This has been updated in the new commit.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,229 @@
...
+
+    conf.get("zookeeper.znode.parent") + "#"
+      conf.get("hbase.zookeeper.quorum") + "#"
+      conf.get("hbase.zookeeper.property.clientPort")
+  }
+}
+
+object SHCCredentialsManager extends Logging {
+
+  def get(sparkConf: SparkConf): SHCCredentialsManager = new SHCCredentialsManager(sparkConf)
```
mridulm: This is creating a new SHCCredentialsManager for each invocation of get - which is for each relation ?
That does not look right - did I miss something ?

weiqingy: `SHCCredentialsManager` should be shared by each relation: f08b44c.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseRelation.scala
weiqingy: No. See [[here](https://github.com/apache/spark/blob/branch-2.1/core/src/main/scala/org/apache/spark/rdd/PairRDDFunctions.scala#L1085)].

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,229 @@
...
+
+  private val tokenUpdater =
+    Executors.newSingleThreadScheduledExecutor(
+      ThreadUtils.namedThreadFactory("HBase Tokens Refresh Thread"))
+
+  private val tokenUpdateRunnable = new Runnable {
+    override def run(): Unit = Utils.logUncaughtExceptions(updateTokensIfRequired())
+  }
+
+  private val isCredentialsManagerEnabled = {
```
mridulm: nit: `isCredentialsManagerEnabled` -> `credentialsManagerEnabled` since it is not a method.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,229 @@
...
+  private val expireTimeFraction = sparkConf.getDouble(SparkHBaseConf.expireTimeFraction, 0.95)
+  private val refreshTimeFraction = sparkConf.getDouble(SparkHBaseConf.refreshTimeFraction, 0.6)
+  private val refreshDurationMins = sparkConf.getInt(SparkHBaseConf.refreshDurationMins, 10)
+
+  private val tokensMap = new mutable.HashMap[String, TokenInfo]
+
+  // We assume token expiration time should be no less than 10 minutes by default.
+  private val nextRefresh = TimeUnit.MINUTES.toMillis(refreshDurationMins)
+
+  private val tokenUpdater =
```
mridulm: nit: `tokenUpdater` -> `tokenUpdateExecutor`

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -0,0 +1,229 @@
...
+
+  private val isCredentialsManagerEnabled = {
+    val isEnabled = sparkConf.getBoolean(SparkHBaseConf.credentialsManagerEnabled, true) &&
+      UserGroupInformation.isSecurityEnabled
+    logInfo(s"SHCCredentialsManager was${if (isEnabled) "" else " not"} enabled.")
+    isEnabled
+  }
+
+  tokenUpdater.scheduleAtFixedRate(
+    tokenUpdateRunnable, nextRefresh, nextRefresh, TimeUnit.MILLISECONDS)
```
mridulm: Do we need to start this if credetials manager is not enabled ?
If it is not required, we can move this and `private val tokenUpdateRunnable` into the block above which sets credentials manager enabled flag.
Note: keep the executor as class instance variable, so that it does not go out of scope.

weiqingy: No, it's not required. I've updated the PR for this. The creation and starting of the executor is not related to the logic of setting credentials manager enabled flag, so I did not put tokenUpdateRunnable inside the block of `credentialsManagerEnabled`.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
weiqingy: Yes. Sure.

weiqingy:
Thanks for the review, @mridulm @jerryshao @bikassaha .

PR #122, opened 2017-04-12
Using principal and keytab to get new tokens

jerryshao:

Wrapping token acquisition with kerberos UGI to make sure to get token with only tgt granted.

## How was this patch tested?

N/A

CC @mridulm @bikassaha @weiqingy

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
@@ -17,7 +17,8 @@
 
 package org.apache.spark.sql.execution.datasources.hbase
 
-import java.io.{DataInputStream, ByteArrayInputStream}
+import java.io.{ByteArrayInputStream, DataInputStream}
+import java.security.PrivilegedExceptionAction
 import java.util.concurrent.{Executors, TimeUnit}
 import java.util.Date
 
@@ -30,6 +31,7 @@ import org.apache.hadoop.hbase.security.token.{AuthenticationTokenIdentifier, To
 import org.apache.hadoop.io.DataOutputBuffer
 import org.apache.hadoop.security.{Credentials, UserGroupInformation}
 import org.apache.hadoop.security.token.{Token, TokenIdentifier}
+
 import org.apache.spark.util.{ThreadUtils, Utils}
 import org.apache.spark.SparkEnv
 import org.apache.spark.sql.execution.datasources.hbase.SHCCredentialsManager._
@@ -47,6 +49,15 @@ final class SHCCredentialsManager private() extends Logging {
   private val expireTimeFraction = sparkConf.getDouble(SparkHBaseConf.expireTimeFraction, 0.95)
   private val refreshTimeFraction = sparkConf.getDouble(SparkHBaseConf.refreshTimeFraction, 0.6)
   private val refreshDurationMins = sparkConf.getInt(SparkHBaseConf.refreshDurationMins, 10)
+
+  private val principal = sparkConf.get(SparkHBaseConf.principal,
+    sparkConf.get("spark.yarn.principal"))
+  private val keytab = sparkConf.get(SparkHBaseConf.keytab,
+    sparkConf.get("spark.yarn.keytab"))
+  require(principal != null, s"neither ${SparkHBaseConf.principal} nor spark.yarn.principal " +
+    s"is configured, this should be configured to make token renewal work")
+  require(keytab != null, s"neither ${SparkHBaseConf.keytab} nor spark.yarn.keytab " +
+    s"is configured, this should be configured to make token renewal work")
 
   private val tokensMap = new mutable.HashMap[String, TokenInfo]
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -48,6 +50,15 @@ final class SHCCredentialsManager private() extends Logging {
   private val refreshTimeFraction = sparkConf.getDouble(SparkHBaseConf.refreshTimeFraction, 0.6)
   private val refreshDurationMins = sparkConf.getInt(SparkHBaseConf.refreshDurationMins, 10)
 
+  private val principal = sparkConf.get(SparkHBaseConf.principal,
+    sparkConf.get("spark.yarn.principal"))
```
mridulm: If both are missing, the exception should indicate both possible option keys. Here, only spark.yarn.principal would be mentioned to the user.

For example, change it to:
```
private val principal = sparkConf.get(SparkHBaseConf.principal,
  sparkConf.getOption("spark.yarn.principal").getOrElse(throw new NoSuchElementException(
    s"${SparkHBaseConf.principal} or spark.yarn.principal must be specified")))
```

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -48,6 +50,15 @@ final class SHCCredentialsManager private() extends Logging {
...
+  private val keytab = sparkConf.get(SparkHBaseConf.keytab,
+    sparkConf.get("spark.yarn.keytab"))
```
mridulm: Same as for principal above.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -48,6 +50,15 @@ final class SHCCredentialsManager private() extends Logging {
...
+  require(principal != null, s"neither ${SparkHBaseConf.principal} nor spark.yarn.principal " +
+    s"is configured, this should be configured to make token renewal work")
+  require(keytab != null, s"neither ${SparkHBaseConf.keytab} nor spark.yarn.keytab " +
+    s"is configured, this should be configured to make token renewal work")
```
mridulm: This block must be done only if `credentialsManagerEnabled` is true - that is fetching pricinpal/keytab and require on them.
If `credentialsManagerEnabled` is false, it is valid for no principal/keytab to be specified (since we never return tokens).

This will be the state for all existing use of shc for example (where there is no dynamic hbase credential acquisition),

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
```diff
@@ -177,12 +188,17 @@ final class SHCCredentialsManager private() extends Logging {
...
+      override def run(): TokenInfo = {
+        val token = TokenUtil.obtainToken(conf)
+        val tokenIdentifier = token.decodeIdentifier()
+        val expireTime = tokenIdentifier.getExpirationDate
+        val issueTime = tokenIdentifier.getIssueDate
+        val refreshTime = getRefreshTime(issueTime, expireTime)
+        new TokenInfo(expireTime, issueTime, refreshTime, conf, token, serializeToken(token))
+      }
+    })
   }
```
mridulm: We need to add token to UGI.currentUser after acquisition.
There is a snippet in `getTokenForCluster` which needs to be moved here - so that the newly acquired token is added to current user's credentials - enabling access to the hbase cluster (and continuing to keep access valid) from driver also.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
jerryshao: Sorry I don't get it. In the `require(...)`, I do indicate bother option keys.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
jerryshao: I see, I will fix it.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/SHCCredentialsManager.scala
jerryshao: get you point now, will fix it.

jerryshao:
Thanks @mridulm for your review, let me merge it. 

@weiqingy hopes we could do some real test on it.

weiqingy:
@mridulm @jerryshao 

The change has been verified in a single secure cluster. The setup of secure multiple clusters has some issue, QE team is working on it. After fixing the issues, will do testing in multiple secure clusters.

The configurations in spark-default.conf:
```
spark.yarn.security.credentials.hbase.enabled false
spark.hbase.connector.security.credentials  ambari-qa-c1@EXAMPLE.COM
spark.hbase.connector.security.keytab  /etc/security/keytabs/smokeuser.headless.keytab
```
Let me know if I missed anything.

PR #193, opened 2017-10-25
add support for multiple versions

btomala:

* extend MAX_VERSIONS flag usage (it was used to set how many versions was fetched only)
* give the ability to create table which holds more than one version
* give the ability to fetch more than one (latest) view of the same entity
* you can set number of versions to fetch

## How was this patch tested?

It contains own suite with few integration tests.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
@@ -143,11 +143,11 @@ private[hbase] class HBaseTableScanRDD(
     }
 
     val ts = valueSeq.foldLeft(Set.empty[Long])((acc, map) => acc ++ map.keySet)
-    //we are loosing duplicate here, because we didn't support passing version (timestamp) to the row
+    // we are loosing duplicate here, because we didn't support passing version (timestamp) to the row
     ts.map(version => {
       keySeq ++ valueSeq.map(_.apply(version)).toMap
     }).map { unioned =>
-      // Return the row ordered by the requested order
+      // return the row ordered by the requested order
       Row.fromSeq(fields.map(unioned.get(_).getOrElse(null)))
     }
   }
@@ -232,9 +232,9 @@ private[hbase] class HBaseTableScanRDD(
   }
 
   /**
-    * Convert result in to list of rows aggregated by timestamp and flat this list into one iterator of rows
-    * This solution stand for fetching more than one version
-    */
+   * Convert result in to list of rows aggregated by timestamp and flat this list into one iterator of rows
+   * This solution stand for fetching more than one version
+   */
   private def toFlattenRowIterator(
       it: Iterator[Result]): Iterator[Row] = {
 
@@ -269,7 +269,7 @@ private[hbase] class HBaseTableScanRDD(
           if(rows.isEmpty) { 
             // If 'requiredColumns' is empty, 'indexedFields' will be empty, which leads to empty 'rows'.
             // This happens when users' query doesn't require Spark/SHC to return any real data from HBase tables,
-            // e.g. dataframe.count().
+            // e.g. dataframe.count()
             Row.fromSeq(Seq.empty)
           } else {
             nextRow()
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
@@ -267,7 +267,10 @@ private[hbase] class HBaseTableScanRDD(
           val r = it.next()
           rows = buildRows(indexedFields, r)
           if(rows.isEmpty) {
-            //why this happened?
+            // This is happened when it performed dataframe.count() (e.g. assert(twoVersions.count() == 3) in the test case).
+            // When performing users' queries, SHC gets the requiredColumns from Spark catalyst engine.
+            // However, when performing count() action, the requiredColumns given by Spark is empty (requiredColumns.length==0).
+            // Actually, that makes sense since what users want is the number of unique rows not the columns' values.
             Row.fromSeq(Seq.empty)
           } else {
             nextRow()
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
@@ -231,6 +231,10 @@ private[hbase] class HBaseTableScanRDD(
     iterator
   }
 
+  /**
+    * Convert result in to list of rows aggregated by timestamp and flat this list into one iterator of rows
+    * This solution stand for fetching more than one version
+    */
   private def toFlattenRowIterator(
       it: Iterator[Result]): Iterator[Row] = {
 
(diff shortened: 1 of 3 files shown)


weiqingy:
Thanks for this PR. I'll review this soon.

weiqingy:
Retriggered travis CI build.

core/src/test/scala/org/apache/spark/sql/MaxVersionsSuite.scala
```diff
@@ -0,0 +1,127 @@
...
+      HBaseRecord(i, "old")
+    }
+    val newData = (0 to 2).map { i =>
+      HBaseRecord(i, "new")
+    }
+    val newestData = (0 to 1).map { i =>
+      HBaseRecord(i, "latest")
+    }
+
+    sc.parallelize(oldestData).toDF.write.options(
```
weiqingy: These four code segments perform the same logic with the difference of just one parameter. When we read the code it takes time to find the difference and maybe it'll be better to create a function and just pass in the different parameters to improve readability.

core/src/test/scala/org/apache/spark/sql/MaxVersionsSuite.scala
```diff
@@ -0,0 +1,127 @@
...
+package org.apache.spark.sql
+
+import org.apache.spark.Logging
+import org.apache.spark.sql.execution.datasources.hbase.{HBaseRelation, HBaseTableCatalog}
+import org.apache.spark.sql.sources.PrunedFilteredScan
+
+
+class MaxVersionsSuite extends SHC with Logging {
+
+  def withCatalog(cat: String): DataFrame = {
```
weiqingy: "withCatalog" function is never used in this test. We can use it to get twoVersions, threeVersions and lastVersions.

weiqingy:
@btomala The PR is useful. Some minor comments.

core/src/test/scala/org/apache/spark/sql/MaxVersionsSuite.scala
```diff
@@ -0,0 +1,127 @@
...
+    sqlContext
+      .read
+      .options(Map(HBaseTableCatalog.tableCatalog->cat))
+      .format("org.apache.spark.sql.execution.datasources.hbase")
+      .load()
+  }
+
+  private def prunedFilterScan(cat: String): PrunedFilteredScan = {
```
weiqingy: This function is never called and useless?

core/src/test/scala/org/apache/spark/sql/MaxVersionsSuite.scala
btomala: Yes you have right I create a function for that

core/src/test/scala/org/apache/spark/sql/MaxVersionsSuite.scala
btomala: Done

core/src/test/scala/org/apache/spark/sql/MaxVersionsSuite.scala
btomala: Removed

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
```diff
@@ -195,6 +231,51 @@ private[hbase] class HBaseTableScanRDD(
...
+        row
+      }
+
+      override def next(): Row = {
+        rowCount += 1
+        if(rows.isEmpty) {
+          val r = it.next()
+          rows = buildRows(indexedFields, r)
+          if(rows.isEmpty) {
+            //why this happened?
```
weiqingy: I found this happened when Result only contains `keySeq` (no `valueSeq` in `buildRows` function). Have you figured out why this happened?

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
```diff
@@ -195,6 +231,51 @@ private[hbase] class HBaseTableScanRDD(
     iterator
   }
```
weiqingy: Could you please add some function comments before `toFlattenRowIterator` (just simply remind people this function is for multiple versions stuff)?

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
btomala: It happened when the list of `fields` is empty, this is the reason why the `valueSeq` is empty. Its construction base on mapping `fields` which is the empty sequence. But I don't know why the fields sometimes are empty. The construction of `HBaseTableScanR DD` sometimes got empty `requiredColumns` array. Do you know when it could happen?

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
btomala: Done

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
weiqingy: I debugged this. This is happened when it performs `dataframe.count()` (e.g. `assert(twoVersions.count() == 3)` in the test case). When performing users' queries, SHC gets the `requiredColumns` from Spark catalyst engine. However, when performing `count()` action, the `requiredColumns` given by Spark is empty (requiredColumns.length==0). Actually, that makes sense since what users want is the number of unique rows not the columns' values. 

Could you please remove the "//why this happened?", and comment what I mentioned above instead? Then I will merge it. 

Also, could you also update the PR for master branch? we can merge it to master/branch-2.2 too.

core/src/test/scala/org/apache/spark/sql/MaxVersionsSuite.scala
```diff
@@ -0,0 +1,108 @@
...
+
+    // Test specific only last versions
+    val lastVersions: DataFrame = withCatalog(catalog, Map.empty)
+
+    val lastRows = lastVersions.take(10)
+    assert(lastRows.size == 3)
+    assert(lastRows.count(_.getString(7).contains("new")) == 1)
+    assert(lastRows.count(_.getString(7).contains("latest")) == 2)
+
+
```
weiqingy: delete the empty lines?

core/src/test/scala/org/apache/spark/sql/MaxVersionsSuite.scala
```diff
@@ -0,0 +1,108 @@
...
+package org.apache.spark.sql
+
+import org.apache.spark.Logging
+import org.apache.spark.sql.execution.datasources.hbase.{HBaseRelation, HBaseTableCatalog}
+
+
+class MaxVersionsSuite extends SHC with Logging {
+
+
```
weiqingy: delete the empty lines?

core/src/test/scala/org/apache/spark/sql/MaxVersionsSuite.scala
btomala: Done

core/src/test/scala/org/apache/spark/sql/MaxVersionsSuite.scala
btomala: Done

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
btomala: Done, also applied all comments into master branch PR

weiqingy:
I updated the code comments a little bit.

weiqingy:
Thanks, @btomala . Merged to branch-1.6.

PR #198, opened 2017-11-14
fix ingestion entities with nulls

btomala:

Give the ability to ingest entities which contain null values.

## How was this patch tested?

I create the unit test which was failing before fix with such exception:
`java.lang.UnsupportedOperationException: PrimitiveType coder: unsupported data type null`

Fix simple omit the null values and didn't try to save it (which is correct, HBase didn't store null)


core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseRelation.scala
```diff
@@ -175,6 +175,42 @@ case class HBaseRelation(
     connection.close()
   }
 
+  private def convertToPut(rkIdxedFields: Seq[(Int, Field)], colsIdxedFields: Array[(Int, Field)])(row: Row) = {
```
weiqingy: The variables `rkIdxedFields`  and `colsIdxedFields` are only used in `convertToPut()`, so if we move `convertToPut()` out of `insert()` function, it's better to move the logic of how to get `rkFields`, `rkIdxedFields`, `colsIdxedFields` into `convertToPut()` too, then we donot need to pass `rkIdxedFields: Seq[(Int, Field)], colsIdxedFields: Array[(Int, Field)])` into `convertToPut()` as parameters.  Otherwise, I'd prefer to `convertToPut()` inside `insert()` function since `convertToPut()` is integral step of `insert()` and only be called in `insert()`.

core/src/test/scala/org/apache/spark/sql/InsertDataSuite.scala
```diff
@@ -0,0 +1,57 @@
...
+    sc.parallelize(data).toDF.write
+      .options(Map(
+        HBaseTableCatalog.newTable -> "5",
+        HBaseTableCatalog.tableCatalog -> cat
+      ))
+      .format("org.apache.spark.sql.execution.datasources.hbase")
+      .save()
+  }
+
+  test("inserting data with null values") {
```
weiqingy: Could you move this test case into DefaultSourceSuite? DefaultSourceSuite is a general test suite for testing reading data, writing data, etc. We donot need to create a new test suite for inserting data.

weiqingy:
Thanks for this PR. Some minor comments.

sachinsga:
when can we expect this pull request to be merged ?

core/src/test/scala/org/apache/spark/sql/InsertDataSuite.scala
btomala: Yes sure I will do that.

btomala:
Hold on, with merging this.The friend of mine told me that HBase is able to store null, I need to check that. Maybe we should all allow ingest nulls instead of skipping them.

btomala:
Or maybe you know what is expected behaviour?

btomala:
Ok, HBase doesn't store nulls. According to HBase documentation, it could store 0 bytes like empty string but not unll. You can merge this PR.

btomala:
@weiqingy Do you have the plan to merge this PR?

weiqingy:
Sorry for the late response. Will check the new commit, and merge it soon (by this week).

core/src/test/scala/org/apache/spark/sql/DefaultSourceSuite.scala
```diff
@@ -78,6 +78,33 @@ class DefaultSourceSuite extends SHC with Logging {
     HBaseRelation(Map(HBaseTableCatalog.tableCatalog->cat),None)(sqlContext)
   }
 
+  def persistDataInHBase(cat: String, data: Seq[HBaseRecord]): Unit = {
```
weiqingy: Currently this function is only used in `test("inserting data with null values") `, but it also can be used in other test cases like ` test("Timestamp semantics") ` which need to save data into Hbase. Could you also refactor those test cases to invoke `persistDataInHBase`?

core/src/test/scala/org/apache/spark/sql/DefaultSourceSuite.scala
```diff
@@ -78,6 +78,33 @@ class DefaultSourceSuite extends SHC with Logging {
...
+    sc.parallelize(data).toDF.write
+      .options(Map(
+        HBaseTableCatalog.newTable -> "5",
+        HBaseTableCatalog.tableCatalog -> cat
+      ))
+      .format("org.apache.spark.sql.execution.datasources.hbase")
+      .save()
+  }
+
+  test("inserting data with null values") {
```
weiqingy: All the test cases in this test suite share the same hbase table "table1", and currently its size is 104. It's better to move `test("inserting data with null values")` below others, and append the new data created in the new test case to the hbase table instead of modifying the existing rows (row1 - row4) , for example:
```
test("inserting data with null values") {
    val originalData = withCatalog(catalog)
    assert(originalData.count() == 104)

    val withNullData = (259 to 260).map(HBaseRecord(_, "").copy(col7 = null))
    val withoutNullData = (261 to 262).map(HBaseRecord(_, "not null"))

    persistDataInHBase(catalog, withNullData ++ withoutNullData)

    val data: DataFrame = withCatalog(catalog)

    assert(data.count() == 108)

    val rows = data.collect()
    assert(rows.count(_.getString(7) == null) == 2)
    assert(rows.count(_.getString(7) != null) == 106)
  }
```

weiqingy:
Hi @btomala There are two minor comments for the test case. I'll merge it after you update it. Thanks.

core/src/test/scala/org/apache/spark/sql/DefaultSourceSuite.scala
btomala: Yes, I understand that but what is a difference between being last and first. This test is first and using rows between 1 to 4, after that there is a test which populates the data and overrides those four rows. I prefer to operate on a clean table, this is the strange convention for me to use one data set in many tests. This is the reason why I put my tests in separate files. Tests should be separated. Each test should have own table or table should be clean before or after each test.

core/src/test/scala/org/apache/spark/sql/DefaultSourceSuite.scala
btomala: done

core/src/test/scala/org/apache/spark/sql/DefaultSourceSuite.scala
weiqingy: I agree with your point to create a new table for the new test case. Currently the table name is hardcoded in [class SHC](https://github.com/hortonworks-spark/shc/blob/master/core/src/test/scala/org/apache/spark/sql/SHC.scala#L56), so if test cases use the same catalog, they will talk with the same hbase table. We can make the table name configurable. How about writing a new function to get different catalogs with different name(e.g  defineCatalog(tablename))?

btomala: I wrote this function and use it in my test. Also, marked old one as deprecated, but doesn't refactor other tests. I think it is out of scope this issue.

weiqingy:
Thanks, @btomala . Merged to mater branch.

rohitsoni11:
@btomala @weiqingy 
Hi, do you know in which version of *s1.6 , this fix is available. As i am using the:1.1.1-1.6-s_2.10 , but facing same issue. Please advise.

btomala:
I don't have idea. I don't work with spark any more.

PR #207, opened 2017-12-21
Feature/dynamic columns

btomala:
# What changes were proposed in this pull request?
Ability to read multiple columns from HBase with the unknown column name.

# How was this patch tested?
It contains own test

# Limitation.
Support only PrimitiveType serialization

# Note
I consider blocking MapType for future use but it shouldn't be a problem because MapType isn't primitive one

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
@@ -22,8 +22,6 @@ package org.apache.spark.sql.execution.datasources.hbase
 
 import java.util.ArrayList
 
-import scala.collection.mutable
-
 import org.apache.hadoop.hbase.CellUtil
 import org.apache.hadoop.hbase.client._
 import org.apache.hadoop.hbase.filter.{Filter => HFilter}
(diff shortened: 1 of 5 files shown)


weiqingy:
Thanks, @btomala . 

cc: @dongjoon-hyun @merlintang

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
```diff
@@ -86,9 +88,33 @@ private[hbase] class HBaseTableScanRDD(
...
   }
 
+  private def extractKey(s: String, r: Regex): Option[String] = s match {
+    case r() => Some("")
+    case r(x) => Some(x)
+    case _ => None
+  }
+
+  private def containsDynamic(dt: DataType): Boolean = dt match {
+    case m: MapType => m.keyType.isInstanceOf[StringType.type]
```
yinthewind: Any special reasons for restricting keyType to be StringType? 
In case qualifier bytes contains 0xFF(or any value bigger than 127) we will got a 0xFFFD for failing to parse it as utf-8 string

btomala: Nope, you can change that. You need be aware that types are a way to identify if it is dynamic column name (StringType as a key in a map) or versions (LongType as a key in a map).

yinthewind:
> Thanks, @btomala .
> 
> cc: @dongjoon-hyun @merlintang

Hi, @weiqingy , is there any blocker for merging this pr?

yinthewind:
Hey @btomala , thanks for your efforts for enabling dynamic columns, do you know why this PR hasn't been merged into master? Anything we can do to help that?

btomala:
Hi @yinthewind, I don't know. I made this change because we need it. And we keep it in private repo because they don't want to merge it. The whole discussion is here. Cheers!

merlintang:
LGTM. if no further comments, I can merge this.

yinthewind:
@merlintang that would be great, thank you!

davidvoit:
Hi,
I think we found the issue with the code, HBaseRelation.RESTRICTIVE -> HBaseRelation.Restrictive.none was missing in the reading code. Now it seems to work.

One thing which would be great is to get some documentation about these RESTRICTIVE modes, what does family mean, what does none means. But we think we get it now.

Thanks and Regards,
David

PR #247, opened 2018-04-19
Add a unit test using a dataset with a field order different than in …

sbarnoud:
…the Avro schema

Added unit test to expose Avro field order problem

## How was this patch tested?

NA


sbarnoud:
May i have a feedback? Travis failed, but this is normal, the test is just intended for that, and a PR propose a fix.

Issue #210, opened 2018-01-22
How to get timestamp column from hbase.


AtulKumVerma:
=> I want to get the timestamp of each row from hbase using SHC. so what i need to specifed in catalog so timestamp column is come with specified column.

=> I have another issue not related to SHC just asking for general purpose. I want see all the registered dataset or dataframe as temporary table or view in Sql context.
I read about it, Catalyst is responsible for maintain one to one mapping between dataframe and its temporary table name.
I want to just list down that all from catalyst.

Your Response highly Appreciate.
Thanks All in Advance.

AtulKumVerma:
@weiqingy 
??

weiqingy:
Hi @AtulKumVerma,

Sorry for the late reply. I somehow missed your question.

_Q: `=> I want to get the timestamp of each row from hbase using SHC. so what i need to specifed in catalog so timestamp column is come with specified column.`_

You may need to add a column for "timestamp" in the catalog when you write data into HBase tables. Just treat "timestamp" the same as other columns.

AtulKumVerma:
let consider i have data like .
ROW                                   COLUMN+CELL                                                                                               
 1                                    column=r:Salary, timestamp=1516691525884, value=50000                                               
 1                                    column=r:designation, timestamp=1516691525884, value=manager                                              
 1                                    column=r:name, timestamp=1516691525884, value=Atul

so i want in hbase read like.
+-----------+---------+----------------+-------+--------------------------+
| rowkey | Salary | designation | name | timestamp              |     
+----------+---------+----------------+--------+--------------------------+
| 1           | 50000 | manager     | XYZ   |  1516691525884    |
+----------+---------+----------------+--------+--------------------------+

so how i get timestamp column in dataframe ?

weiqingy:
Currently SHC doesnot support this case. It can read out the value of each column only. You may need to customize SHC [here](https://github.com/hortonworks-spark/shc/blob/master/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala#L291) and [here](https://github.com/hortonworks-spark/shc/blob/master/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala#L318) to support this.

AtulKumVerma:
ok

amruth-quaero:
Hi @weiqingy, 
Could you please review this PR. I need this change for one of the solutions.

PR #297, opened 2018-12-13
added HBase timestamp support

sbarnoud:
https://github.com/hortonworks-spark/shc/issues/210

Added support of HBase timestamp.

## How was this patch tested?

New TU. Added control in TU that "reserved" column families (rowkey and timestamp) are not created.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
@@ -56,7 +56,7 @@ private[hbase] class HBaseTableScanRDD(
     requiredColumns: Array[String],
     filters: Array[Filter]) extends RDD[Row](relation.sqlContext.sparkContext, Nil) {
   val outputs = StructType(requiredColumns.map(relation.schema(_))).toAttributes
-  val columnFields = relation.splitRowKeyColumns(requiredColumns)._2
+  val columnFields = relation.splitReservedColumns(requiredColumns)._2
   private def sparkConf = SparkEnv.get.conf
   val serializedToken = relation.serializedToken
 
@@ -125,6 +125,15 @@ private[hbase] class HBaseTableScanRDD(
       }
     }
 
+    val tsSeq = {
+      if (relation.catalog.ts.isPresent) {
+        val f = relation.catalog.getTimestamp.head
+        Seq((f, result.rawCells()(0).getTimestamp)).toMap
+      } else {
+        Seq.empty
+      }
+    }
+
     import scala.collection.JavaConverters.mapAsScalaMapConverter
     val scalaMap = result.getMap.asScala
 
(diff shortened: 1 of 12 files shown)


core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
```diff
@@ -228,7 +246,16 @@ private[hbase] class HBaseTableScanRDD(
       }
     }
 
-    val valueSeq: Seq[Map[Long, (Field, Any)]] = fields.filter(!_.isRowKey).map { x =>
+    val tsSeq = {
+      if (relation.catalog.ts.isPresent) {
+        val f = relation.catalog.getTimestamp.head
+        Seq((f, result.rawCells()(0).getTimestamp)).toMap
```
regata: @sbarnoud when there are multiple cells in a single column, each cell will have different timestamp. However, this code only reads the latest cell. Is this intended behavior?

regata: plus `tsSeq` is not used inside `buildRows`

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableScan.scala
```diff
@@ -228,7 +246,16 @@ private[hbase] class HBaseTableScanRDD(
...
+      } else {
+        Seq.empty
+      }
+    }
+
+    val valueSeq: Seq[Map[Long, (Field, Any)]] = fields.filter(c => HBaseTableCatalog.isNotReserved(c.cf)).map { x =>
```
regata: how about this?

```scala
    val tsField = relation.catalog.getTimestamp.head // ADDED
    val valueSeq: Seq[Map[Long, (Field, Any)]] = fields.filter(c => HBaseTableCatalog.isNotReserved(c.cf)).map { x =>
      import scala.collection.JavaConverters.asScalaBufferConverter
      val dataType = SHCDataTypeFactory.create(x)
      val kvs = result.getColumnCells(
        relation.catalog.shcTableCoder.toBytes(x.cf),
        relation.catalog.shcTableCoder.toBytes(x.col)).asScala

      kvs.map(kv => {
        val v = CellUtil.cloneValue(kv)
        (kv.getTimestamp, x -> dataType.fromBytes(v))
      }).toMap.withDefaultValue(x -> null)
    }

    val ts = valueSeq.foldLeft(Set.empty[Long])((acc, map) => acc ++ map.keySet)
    //we are loosing duplicate here, because we didn't support passing version (timestamp) to the row
    ts.map(version => {
      val tsSeq = Seq((tsField, version)).toMap // ADDED
      keySeq ++ tsSeq ++ valueSeq.map(_.apply(version)).toMap // MODIFIED
    }).map { unioned =>
      // Return the row ordered by the requested order
      Row.fromSeq(fields.map(unioned.get(_).getOrElse(null)))
    }
```

The only extra change that is required is to specify timestamp column when `mergeToLatest` is false

sbarnoud: My bad.In my use case, i have a single version, and i forget this point.
Yes, we should have the ts corresponding to the version!

Issue #330, opened 2019-12-18
Numeric Keys are Not Split Properly


davidov541:
If you have keys that start exclusively with numeric values, all rows will be put into a single region. That region will be split as necessary based on the settings in HBase, but the other regions will remain empty. For example, if I request 100 regions, it will create 100 regions between "aaaaaa" and "zzzzzz". If I then add 1 million rows with numeric keys, all of them will go into the first region. If I add more rows, then the HBase cluster may decide to split the first region, but the other regions will remain empty. 

I've included some source code below that I've written which demonstrates this issue easily. Increasing the number of samples allows you to see the new regions were created, but even at 1 million, you wouldn't expect the minimum and maximum keys to be in the same region if you asked for 100 regions.

```
val numValues = 1000000
val startValue = 48000000

val df = spark.range(startValue, startValue + numValues).select($"id".cast("string") as "pat_id", lit("RUL_ID_VALUE") as "rul_id", lit("ROW_KEY_VALUE") as "rowkey")

/* Cast DataTypes */

val cir_fitered_data_DF = df.select(
  $"pat_id" as "fullrowkey",
  $"rul_id",
  $"rowkey" as "ROW_KEY");

//Schema to load into HBase perf table

def mbr_cir_load_hbase_catalog =
  s"""{
     |"table":{"namespace": "u405317_sandbox", "name":"PerfTest"},
     |"rowkey":"key",
     |"columns":{
     |"fullrowkey":{"cf":"rowkey", "col":"key", "type":"string"},
     |"rul_id":{"cf":"P1", "col":"rul_id", "type":"string"},
     |"ROW_KEY":{"cf":"P1", "col":"rowkey", "type":"string"}
     |}
     |}""".stripMargin

cir_fitered_data_DF.show(false);

cir_fitered_data_DF.write.options(Map(HBaseTableCatalog.tableCatalog -> mbr_cir_load_hbase_catalog, HBaseTableCatalog.newTable -> "100")).format("org.apache.spark.sql.execution.datasources.hbase").save();
```

PR #331, opened 2019-12-18
[SHC-330] Adding ability to set first and last splits, so that non-alphabetic splits can be easily supported

davidov541:

Adds two properties that allow the first and last split to be set by the user. This allows the use of more complex split preferences, as well as enables scenarios such as numeric-only keys, which were underserved in earlier versions.

## How was this patch tested?

Two unit tests were added. One tests that for alphabetic keys, the rows are dispersed between regions correctly. The other uses numeric keys and sets the first and last splits accordingly. This also checks that the rows have been dispersed appropriately across regions.


davidov541:
Fixes issue https://github.com/hortonworks-spark/shc/issues/330

peter-toth:
Looks good to me, @dongjoon-hyun could you pleas take a look as well?

peter-toth (commented):
Why do we need `.github/workflows/maven.yml` in this PR?

davidov541:
@peter-toth : We don't need maven.yml. I suspect that was added automatically when I was trying to set up a maven repository to help the customer, and didn't realize it was included as part of the PR. It's now been removed.

cricket007:
Might want to add `.github/workflows` to the gitignore until if/when TravisCI is migrated to Github Actions.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
```diff
@@ -250,6 +251,8 @@ object HBaseTableCatalog {
   val tableCoder = "tableCoder"
   // The version number of catalog
   val cVersion = "version"
+  val minRegionSplit = "minregionsplit"
+  val maxRegionSplit = "maxregionsplit"
```
cricket007: Not familar with where `parameters` is defined but I would assume these strings are  camelCase

davidov541:
@cricket007 Fixed the internal names. Thanks!

gaborgsomogyi (approved):
LGTM.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
```diff
@@ -250,6 +251,8 @@ object HBaseTableCatalog {
   val tableCoder = "tableCoder"
   // The version number of catalog
   val cVersion = "version"
+  val minRegionSplit = "minRegionSplit"
+  val maxRegionSplit = "maxRegionSplit"
```
joshelser: While I can appreciate these names and what you're trying to allow the user to provide, they aren't using HBase nomenclature well.

I think the names `minTableSplitPoint` and `maxTableSplitPoint` would get the point across better what they're doing.

joshelser (commented):
Thanks for putting this together, David.

Having read this a few times, the logic on the number of regions which will be created by SHC is a little... obtuse, to say the least. Best as I can see, SHC is doing everything correctly to avoid an error, but table distribution (the split points applied) might be surprising to a user.

They will get, at a minimum, 4 regions because a minimum of three split-points are chosen: the "min", the "max", and one split point lexicographically equidistant between the min and the max (e.g. if `min=A` and `max=Z`, we'd split between `M` and `N`). If the user requests more regions, we will generate more split points between the min and the max.

That's all to say, I think the user documentation could benefit from explaining this somewhere, rather than let the user figure this out via experimentation :). Maybe we can add something to the README as a follow-up?

davidov541:
Thanks, @joshelser. I'm working on a rename of the parameter to match your request. I think that seems reasonable. Not being an HBase expert myself, I wasn't clear on the difference there, but I get it.

As for the README, I agree. In general, the README could use a lot of work. My understanding is that this project is EOL essentially. If so, is it still worth updating the README at this point in time? If the PR will be reviewed and accepted in a reasonable amount of time, I'd be down to update the README at the very least for this portion.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
davidov541: Change has been made. Thanks @joshelser!

peter-toth (approved):
LGTM. Thanks.

@joshelser please review before I merge it.

core/src/main/scala/org/apache/spark/sql/execution/datasources/hbase/HBaseTableCatalog.scala
peter-toth: @joshelser, these parameters are called now `minRegionSplitPoint` and `maxRegionSplitPoint`. Do you think `minTableSplitPoint` and `maxTableSplitPoint` would be better names?

joshelser:
> please review before I merge it.

Looks like the configuration names changed, but not to what I had suggested: `minRegionSplitPoint` and `maxRegionSplitPoint`. I suggested `minTableSplitPoint` and `maxTableSplitPoint`. We're adding splits to a table which creates more regions. We're not adding split points to Regions. Sorry for being pedantic.

> My understanding is that this project is EOL essentially. If so, is it still worth updating the README at this point in time

Great question that I don't have a good answer for :). Usually I follow a rule of "if I come back to something for a second time, I should fix the documentation because there will be a third time." Ultimately it's up to you. Hopefully we have the right folks with the right permissions to make changes going forward.

davidov541:
@joshelser : Don't worry about it. It's my fault for not paying enough attention. I'll get on that and update the PR in the next week or two.

I'll go ahead and make the change we've discussed to the README and submit a PR to everyone at some point in the next week.

joshelser (approved):
Thanks for fixing, David!

peter-toth:
Thanks @davidov541 for your work. I've merged it to `master`.

